PR #1, opened 2019-11-04
XX handshake

morrigan:


src/xx.ts
```diff
@@ -0,0 +1,138 @@
...
+    const cs = this.initializeKey(emptyKey);
+
+    return { cs, ck, h };
+  }
+
+  async hashProtocolName(protocolName: string) : Promise<bytes32> {
+    if (protocolName.length <= 32) {
+      return new Promise(resolve => {
+        const h = new Buffer(32);
+        h.write(protocolName);
```
mpetrunic: you need to convert string to bytes using utf-8 encoding (go by default converts string to bytes using utf-8). Add explicit converting to buffer using utf 8 encoding

src/xx.ts
```diff
@@ -0,0 +1,138 @@
...
+        resolve(h)
+      });
+    } else {
+      return await this.getHash(Buffer.from(protocolName), new Buffer([]));
```
mpetrunic: Buffer.from(protocolName, 'utf-8'). It's probably default, but it's nice to keep it explicit

src/xx.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  async mixHash(ss: SymmetricState, data: bytes) {
+    ss.h = await this.getHash(ss.h, data);
+  }
+
+  async getHash(a: bytes, b: bytes) : Promise<bytes32> {
+    return await crypto.hmac.create('sha256', Buffer.from([...a, ...b]))
```
mpetrunic: blake2s.Sum256(append(a, b...)) !== sha256

https://github.com/bcoin-org/bcrypto/blob/master/lib/js/blake2s.js

Sum256 means bits value is 256 (default for blake2s in that library I think)

src/xx.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  async initSession(initiator: boolean, prologue: bytes32[], s: KeyPair, rs: bytes32) : Promise<NoiseSession> {
+    let session: NoiseSession;
+    const psk = emptyKey;
```
mpetrunic: careful here. I would rather make emptyKey to be function that generates Buffer.alloc(32). As functions here will modify empty key variable (Buffer is object passed as reference and not primitive)

src/xx.ts
```diff
@@ -0,0 +1,138 @@
...
+    const ss = await this.initializeSymmetric(name);
+    await this.mixHash(ss, prologue);
+
+    return {ss, s, e, rs, re, psk};
+  }
+
+  async initializeResponder(prologue: bytes32, s: KeyPair, rs: bytes32, psk: bytes32) : Promise<HandshakeState> {
+    const e: KeyPair;
+    const re: bytes32;
+    const name = "Noise_XX_25519_ChaChaPoly_SHA256";
```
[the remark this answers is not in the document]
morrigan: resolving as per spec https://github.com/libp2p/specs/tree/master/noise#valid-noise-protocol-names

src/xx.ts
morrigan: closing in favor of spec

src/xx.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    return {
+      hs,
+      i: initiator,
+      mc: 0
+    };
+  }
+
+  public async sendMessage(session: NoiseSession, message: bytes) : Promise<MessageBuffer> {
+    let messageBuffer: MessageBuffer = {} as MessageBuffer;
```
mpetrunic: you don''t need to initialize it

src/xx.ts
```diff
@@ -0,0 +1,302 @@
...
+    if (session.mc === 0) {
```
mpetrunic: mc is uint64, so you need to compare it like `session.mc.eqn(0)`

src/xx.ts
```diff
@@ -0,0 +1,302 @@
...
+      messageBuffer = await this.writeMessageA(session.hs, message);
+    } else if (session.mc === 1) {
```
mpetrunic: same as above

src/xx.ts
```diff
@@ -0,0 +1,302 @@
...
+      messageBuffer = await this.writeMessageB(session.hs, message);
+    } else if (session.mc === 2) {
```
mpetrunic: same as above

src/xx.ts
```diff
@@ -0,0 +1,302 @@
...
+      const { h, messageBuffer, cs1, cs2 } = await this.writeMessageC(session.hs, message);
+      session.h = h;
+      session.cs1 = cs1;
+      session.cs2 = cs2;
+    } else if (session.mc > 2) {
```
mpetrunic: session.mc.gtn(2)

src/xx.ts
```diff
@@ -0,0 +1,302 @@
...
+          throw new Error("CS2 (cipher state) is not defined")
+        }
+
+        messageBuffer = await this.writeMessageRegular(session.cs2, message);
+      }
+    } else {
+      throw new Error("Session invalid.")
+    }
+
+    session.mc++;
```
mpetrunic: session.mc.addn(1)

src/xx.ts
```diff
@@ -0,0 +1,302 @@
...
+    if (initiator) {
+      hs = await this.initializeInitiator(prologue, s, rs, psk);
+    } else {
+      hs = await this.initializeResponder(prologue, s, rs, psk);
+    }
+
+    return {
+      hs,
+      i: initiator,
+      mc: 0
```
mpetrunic: new BN(0)

src/xx.ts
```diff
@@ -0,0 +1,302 @@
...
+    const cs2 = this.initializeKey(tempk2);
+
+    return { cs1, cs2 };
+  }
+
+  private async writeMessageA(hs: HandshakeState, payload: bytes) : Promise<MessageBuffer> {
+    let ns = Buffer.alloc(0);
+    hs.e = await this.generateKeypair();
+    const ne = hs.e.publicKey;
+    await this.mixHash(hs.ss, ne);
```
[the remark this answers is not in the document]
morrigan: Will resolve this later as it will be easier for me right now to see the diff.

src/xx.ts
```diff
@@ -158,20 +160,20 @@ export class XXHandshake {
...
     }
   }
 
-  private getHkdf(ck: bytes32, ikm: bytes) : Array<Buffer> {
+  public getHkdf(ck: bytes32, ikm: bytes) : Hkdf {
     const info = Buffer.alloc(0);
-    const prk = HKDF.extract('SHA256', ikm, ck);
-    const okm = HKDF.expand('SHA256', prk, info, ikm.length);
+    const prk = HKDF.extract(SHA256, ikm, ck);
+    const okm = HKDF.expand(SHA256, prk, info, 96);
```
mpetrunic: i would add comment explaining 96 here

morrigan:
@mpetrunic you can check now :) also, look into test if the assertion of right things is good

package.json
```diff
@@ -34,6 +38,15 @@
       "@babel/preset-typescript"
     ],
     "plugins": [
+      "@babel/plugin-transform-runtime",
```
mpetrunic: do we really need this?

It requires babel runtime as prod dependency

payload.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+package pb;
+
+message NoiseHandshakePayload {
+    bytes libp2p_key = 1;
+    bytes noise_static_key_signature = 2;
+    bytes libp2p_data = 3;
+    bytes libp2p_data_signature = 4;
+}
```
mpetrunic: put it in separate directory

test/xx.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const [k1, k2, k3] = xx.getHkdf(ck, ikm);
+    expect(k1.toString('hex')).to.equal('cc5659adff12714982f806e2477a8d5ddd071def4c29bb38777b7e37046f6914');
+    expect(k2.toString('hex')).to.equal('a16ada915e551ab623f38be674bb4ef15d428ae9d80688899c9ef9b62ef208fa');
+    expect(k3.toString('hex')).to.equal('ff67bf9727e31b06efc203907e6786667d2c7a74ac412b4d31a80ba3fd766f68');
+  })
+
+  async function generateEd25519Keys() {
+    return await crypto.keys.generateKeyPair('ed25519');
+  }
```
mpetrunic: extract util test functions to separate file

test/xx.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    const payloadProtoBuf = await protobuf.load("payload.proto");
+    const NoiseHandshakePayload = payloadProtoBuf.lookupType("pb.NoiseHandshakePayload");
+    const payloadInit = NoiseHandshakePayload.create({
+      libp2pKey: libp2pInitKeys.bytes,
+      noiseStaticKeySignature: initSignedPayload,
+    });
+    const payloadInitEnc = NoiseHandshakePayload.encode(payloadInit).finish();
+
+    // initiator sends message
+    const message = Buffer.concat([Buffer.alloc(0), payloadInitEnc]);
```
mpetrunic: this doesn't do anything though haha

test/xx.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      libp2pKey: libp2pRespKeys.bytes,
+      noiseStaticKeySignature: respSignedPayload,
+    });
+    const payloadRespEnc = NoiseHandshakePayload.encode(payloadResp).finish();
+
+    const message1 = Buffer.concat([message, payloadRespEnc]);
+    const messageBuffer2 = await xx.sendMessage(nsResp, message1);
+
+    expect(messageBuffer2.ne.length).not.equal(0);
+    expect(messageBuffer2.ns.length).not.equal(0);
```
mpetrunic: aren't you suppose to check that you received sent message?

PR #3, opened 2019-11-12
New interface with XX handshake

morrigan:
Interface according to: https://github.com/libp2p/js-interfaces/tree/master/src/crypto

and integration of handshake using XX


src/handshake.ts
```diff
@@ -0,0 +1,19 @@
+import { bytes, bytes32 } from "./types/basic";
+import { NoiseSession, XXHandshake } from "./xx";
+import { KeyPair } from "./types/libp2p";
+
+export class Handshake {
+  static async runXX(
```
mpetrunic: I would convert this to constructor and have methods like init/exchange/finish like in secio

mpetrunic: later on it would allow us to combine multiple handshakes

src/index.ts
```diff
@@ -1,8 +1,5 @@
-import { encrypt } from './encrypt';
-
-const tag = '/noise';
+import { Noise } from './noise';
 
 export {
-  tag,
-  encrypt,
+  Noise,
```
mpetrunic: export * from "./noise"

src/noise.ts
```diff
@@ -0,0 +1,81 @@
...
+        privateKey: staticNoiseKey,
+        publicKey,
+      }
+    } else {
+      // todo: generate new static key
+    }
+  }
+
+  public protocol() {
+    return '/noise';
```
mpetrunic: I think this should be `public protocol = "/noise";`

src/types/libp2p.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+
+// Also seen as Pair
+export type Stream = {
+  sink(source: Iterable<any>),
+  source: Object,
+}
+
+export type Duplex = [Stream, Stream];
```
mpetrunic: this belongs in it-peer or something so you have to do 

```javascript
declare module "it-peer" {
    export type Duplex = [Stream, Stream];
}
```

src/types/libp2p.ts
```diff
@@ -0,0 +1,60 @@
...
+  stats: ConnectionStats,
+
+  streams(): Duplex,
+  addStream(muxedStream: any) : Stream,
+}
+
+export interface NoiseConnection {
+  remoteEarlyData?(): bytes,
+  secureOutbound(insecure: InsecureConnection, remotePeer: PeerId): Promise<SecureConnection>,
+  secureInbound(insecure: InsecureConnection): Promise<SecureConnection>,
```
mpetrunic: I don't think that's correct types.
Secio returns just {conn, remotePeer}

morrigan: Mhm, will update according to https://github.com/libp2p/js-interfaces/tree/master/src/crypto#ap

morrigan:
@mpetrunic I think we can merge so it doesn't get long. Still missing encryption/decryption after handshake

src/xx.ts
```diff
@@ -272,6 +277,7 @@ export class XXHandshake {
   }
 
   private async readMessageA(hs: HandshakeState, message: MessageBuffer) : Promise<bytes> {
+    console.log("publci key: ", message.ne)
```
mpetrunic: remove console. You should probably add npm package `debug` and log stuff (separate PR probably)

PR #4, opened 2019-11-28
Create encrypted streams

morrigan:
Work done: 
 - Creating encrypted connection
 - Fix handling identity and static keys

Still missing validation of signature - another PR.


src/handshake.ts
```diff
@@ -1,89 +1,125 @@
...
 
   // stage 0
-  async propose(isInitiator: boolean, earlyData?: bytes): Promise<NoiseSession> {
-    const ns = await this.xx.initSession(isInitiator, this.prologue, this.staticKeys, this.remotePublicKey);
-
-    if (isInitiator) {
-      const signedPayload = signPayload(this.staticKeys.privateKey, getHandshakePayload(this.staticKeys.publicKey));
+  async propose(earlyData?: bytes): Promise<void> {
+    if (this.isInitiator) {
+      const signedPayload = signPayload(this.libp2pPrivateKey, getHandshakePayload(this.staticKeys.publicKey));
```
mpetrunic: add log that it's proposing

mpetrunic: generally try to log, if I'm initiator and when I start something or I received something and when I finish something

src/handshake.ts
```diff
@@ -1,89 +1,125 @@
...
+      logger('Stage 2 - Initiator sent message.');
     } else {
       const receivedMessageBuffer = (await this.connection.readLP()).slice();
-      const plaintext = await this.xx.recvMessage(session, decodeMessageBuffer(receivedMessageBuffer));
+      const plaintext = await this.xx.recvMessage(this.session, decodeMessageBuffer(receivedMessageBuffer));
+      logger('Stage 2 - Responder received the message, finished handshake.')
+    }
+  }
+
+  encrypt(plaintext: bytes, session: NoiseSession): bytes {
```
mpetrunic: I'm pretty sure you won't get buffer from stream, it will most likely be Uint8Array so it might be good idea to adjust types and wrap it here using Buffer.from

src/handshake.ts
morrigan: agree, haven't finished with the logging

src/handshake.ts
morrigan: you mean to accept anything and explicitly convert?

mpetrunic: not anything, plaintext won't be bytes type, it will be Uint8Array

PR #9, opened 2019-12-24
Fix sending payloads larger than 65,519 bytes

### LINKED ISSUE #7: plaintext payloads larger than 65,519 bytes will fail to decrypt
Because the max length of a Noise message is 65,535 bytes, including 16 bytes of AEAD authentication data, plaintext messages larger than 65,519 bytes need to be split up into smaller chunks before encrypting.  Otherwise, the encrypted payloads will end up being prefixed with the wrong length (since the length-prefixes max out at 65,535) and decryption will fail.

https://github.com/libp2p/go-libp2p-noise/pull/10 fixes this same issue in the go implementation, if that's helpful for reference.

morrigan:
Resolves #7

### DIFF
--- a/src/crypto.ts
+++ b/src/crypto.ts
@@ -35,8 +35,17 @@ export function decryptStream(handshake: Handshake): ReturnEncryptionWrapper {
   return async function * (source) {
     for await (const chunk of source) {
       const chunkBuffer = Buffer.from(chunk);
-      const decrypted = await handshake.decrypt(chunkBuffer, handshake.session);
-      yield decrypted
+
+      for (let i = 0; i < chunkBuffer.length; i += maxPlaintextLength) {
+        let end = i + maxPlaintextLength;
+        if (end > chunkBuffer.length) {
+          end = chunkBuffer.length;
+        }
+
+        const chunk = chunkBuffer.slice(i, end);
+        const decrypted = await handshake.decrypt(chunk, handshake.session);
+        yield decrypted;
+      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/crypto.ts
```diff
@@ -1,17 +1,27 @@
...
   (source: Iterable<Uint8Array>): AsyncIterableIterator<Uint8Array>;
 }
 
+const maxPlaintextLength = 65519;
+
 // Returns generator that encrypts payload from the user
 export function encryptStream(handshake: Handshake): ReturnEncryptionWrapper {
   return async function * (source) {
     for await (const chunk of source) {
       const chunkBuffer = Buffer.from(chunk);
```
mpetrunic: I'm pretty sure this is creating large overhead because it's allocating new memory and copy chunk data.
I think following code reuses existing memory:
```javascript
Buffer.from(chunk.buffer, chunk.byteOffset, chunk.length)
```

Ref: https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length

src/crypto.ts
```diff
@@ -22,8 +32,17 @@ export function decryptStream(handshake: Handshake): ReturnEncryptionWrapper {
...
```
mpetrunic: same as above

test/utils.ts
```diff
@@ -11,3 +12,15 @@ export function getKeyPairFromPeerId(peerId: PeerId): KeyPair {
     publicKey: peerId.marshalPubKey(),
   }
 }
+
+export function getRandomBuffer(size: number) : bytes {
```
mpetrunic: exists in bcrypto

PR #10, opened 2019-12-25
IK handshake

morrigan:
- moved common functions to separate abstract class
- ported IK handshake 

Note: I think better file naming should be made... :dagger:

### DIFF
--- a/src/handshakes/abstract-handshake.ts
+++ b/src/handshakes/abstract-handshake.ts
@@ -5,7 +5,7 @@ import {bytes, bytes32, uint32} from "../@types/basic";
 import {CipherState, MessageBuffer, SymmetricState} from "../@types/handshake";
 import {getHkdf} from "../utils";
 
-const minNonce = 0;
+export const MIN_NONCE = 0;
 
 export abstract class AbstractHandshake {
   public encryptWithAd(cs: CipherState, ad: bytes, plaintext: bytes): bytes {
@@ -122,7 +122,7 @@ export abstract class AbstractHandshake {
   }
 
   protected initializeKey(k: bytes32): CipherState {
-    const n = minNonce;
+    const n = MIN_NONCE;
     return { k, n };
   }
 
--- a/src/handshakes/abstract-handshake.ts
+++ b/src/handshakes/abstract-handshake.ts
@@ -8,6 +8,39 @@ import {getHkdf} from "../utils";
 export class AbstractHandshake {
   protected minNonce = 0;
 
+  public encryptWithAd(cs: CipherState, ad: bytes, plaintext: bytes): bytes {
+    const e = this.encrypt(cs.k, cs.n, ad, plaintext);
+    this.setNonce(cs, this.incrementNonce(cs.n));
+
+    return e;
+  }
+
+  public decryptWithAd(cs: CipherState, ad: bytes, ciphertext: bytes): bytes {
+    const plaintext = this.decrypt(cs.k, cs.n, ad, ciphertext);
+    this.setNonce(cs, this.incrementNonce(cs.n));
+
+    return plaintext;
+  }
+
+
+  // Cipher state related
+  protected hasKey(cs: CipherState): boolean {
+    return !this.isEmptyKey(cs.k);
+  }
+
+  protected setNonce(cs: CipherState, nonce: uint32): void {
+    cs.n = nonce;
+  }
+
+  protected createEmptyKey(): bytes32 {
+    return Buffer.alloc(32);
+  }
+
+  protected isEmptyKey(k: bytes32): boolean {
+    const emptyKey = this.createEmptyKey();
+    return emptyKey.equals(k);
+  }
+
   protected incrementNonce(n: uint32): uint32 {
     return n + 1;
   }
@@ -31,6 +64,41 @@ export class AbstractHandshake {
     return plaintext;
   }
 
+  protected encryptAndHash(ss: SymmetricState, plaintext: bytes): bytes {
+    let ciphertext;
+    if (this.hasKey(ss.cs)) {
+      ciphertext = this.encryptWithAd(ss.cs, ss.h, plaintext);
+    } else {
+      ciphertext = plaintext;
+    }
+
+    this.mixHash(ss, ciphertext);
+    return ciphertext;
+  }
+
+  protected decrypt(k: bytes32, n: uint32, ad: bytes, ciphertext: bytes): bytes {
+    const nonce = this.nonceToBytes(n);
+    const ctx = new AEAD();
+
+    ctx.init(k, nonce);
+    ctx.aad(ad);
+    ctx.decrypt(ciphertext);
+
+    // Decryption is done on the sent reference
+    return ciphertext;
+  }
+
+  protected decryptAndHash(ss: SymmetricState, ciphertext: bytes): bytes {
+    let plaintext;
+    if (this.hasKey(ss.cs)) {
+      plaintext = this.decryptWithAd(ss.cs, ss.h, ciphertext);
+    } else {
+      plaintext = ciphertext;
+    }
+
+    this.mixHash(ss, ciphertext);
+    return plaintext;
+  }
 
   protected dh(privateKey: bytes32, publicKey: bytes32): bytes32 {
     const derived = x25519.derive(publicKey, privateKey);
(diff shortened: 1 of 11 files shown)


src/handshakes/abstract-handshake.ts
```diff
@@ -0,0 +1,171 @@
+import {Buffer} from "buffer";
+import { AEAD, x25519, HKDF, SHA256 } from 'bcrypto';
+
+import {bytes, bytes32, uint32} from "../@types/basic";
+import {CipherState, MessageBuffer, SymmetricState} from "../@types/handshake";
+import {getHkdf} from "../utils";
+
+export abstract class AbstractHandshake {
```
mpetrunic: Those all methods are pure, meaning they don't need to be in class, they can be in file called handshake-utils.

src/handshakes/abstract-handshake.ts
```diff
@@ -0,0 +1,171 @@
...
+  protected minNonce = 0;
```
mpetrunic: this is constant. Using it like this is misleading

src/handshakes/ik.ts
```diff
@@ -0,0 +1,172 @@
...
+        messageBuffer = this.writeMessageRegular(session.cs1, message);
+      } else {
+        if (!session.cs2) {
+          throw new Error("CS2 (cipher state) is not defined")
+        }
+
+        messageBuffer = this.writeMessageRegular(session.cs2, message);
+      }
+    } else {
+      throw new Error("Session invalid.")
```
mpetrunic: go version doesn't have this else clause

src/handshakes/ik.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    session.mc = session.mc.add(new BN(1));
```
mpetrunic: addn?

src/handshakes/ik.ts
```diff
@@ -0,0 +1,172 @@
...
+      session.h = h;
+      session.cs1 = cs1;
+      session.cs2 = cs2;
+    } else if (session.mc.gtn(1)) {
+      if (session.i) {
+        if (!session.cs1) {
+          throw new Error("CS1 (cipher state) is not defined")
+        }
+
+        messageBuffer = this.writeMessageRegular(session.cs1, message);
```
mpetrunic: Aren't you suppose to clear session.hs here?

src/handshakes/ik.ts
```diff
@@ -0,0 +1,172 @@
...
+  public recvMessage(session: NoiseSession, message: MessageBuffer): bytes {
+    let plaintext: bytes;
+    if (session.mc.eqn(0)) {
+      plaintext = this.readMessageA(session.hs, message);
+    } else if (session.mc.eqn(1)) {
+      const { plaintext: pt, h, cs1, cs2 } = this.readMessageB(session.hs, message);
+      plaintext = pt;
+      session.h = h;
+      session.cs1 = cs1;
+      session.cs2 = cs2;
```
mpetrunic: Aren't you suppose to clear session.hs here?

src/handshakes/ik.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  private readMessageA(hs: HandshakeState, message: MessageBuffer): bytes {
+    if (x25519.publicKeyVerify(message.ne)) {
+      hs.re = message.ne;
+    }
+
+    this.mixHash(hs.ss, hs.re);
+    this.mixKey(hs.ss, this.dh(hs.s.privateKey, hs.re));
+    const ns = this.decryptAndHash(hs.ss, message.ns);
```
mpetrunic: You are propagating error to  whoever is using handshake? 

Asking because go code is using 'valid' flags and has some conditions based on it?!

src/handshakes/ik.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (ns.length === 32 && x25519.publicKeyVerify(message.ns)) {
```
mpetrunic: you should probably extract public key verification to separate method.

src/handshakes/ik.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  private readMessageB(hs: HandshakeState, message: MessageBuffer) {
+    if (x25519.publicKeyVerify(message.ne)) {
+      hs.re = message.ne;
+    }
+
+    this.mixHash(hs.ss, hs.re);
+    if (!hs.e) {
+      throw new Error("Handshake state should contain ephemeral key by now.");
```
mpetrunic: didn't see this error in go impl?

test/utils.ts
```diff
@@ -11,3 +12,15 @@ export function getKeyPairFromPeerId(peerId: PeerId): KeyPair {
     publicKey: peerId.marshalPubKey(),
   }
 }
+
+export function getRandomBuffer(size: number) : bytes {
```
mpetrunic: bcrypto has that method

src/crypto.ts
```diff
@@ -1,17 +1,27 @@
-import { Handshake } from "./handshake";
 import { Buffer } from "buffer";
+import { Handshake } from "./handshake";
 
 interface ReturnEncryptionWrapper {
   (source: Iterable<Uint8Array>): AsyncIterableIterator<Uint8Array>;
 }
 
+const maxPlaintextLength = 65519;
```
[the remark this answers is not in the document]
morrigan: Yep, check it first

src/handshakes/abstract-handshake.ts
morrigan: This is true but I am not sure if it's the right approach. Made more sense like this to me

src/handshakes/ik.ts
morrigan: They are not 100% same language... :D Should never end here anyway.

src/handshakes/ik.ts
morrigan: I am not sure if I am supposed to but I guess I could as it is not required anymore.

src/handshakes/ik.ts
morrigan: ?

src/handshakes/ik.ts
morrigan: They have different interface that includes those valid flags. In JS, if invalid, it's only throwing errors so I guess more try/catch blocks should be involved to substitute it.

src/handshakes/ik.ts
morrigan: Of course not because we are using typescript :D Must check if existing.

src/handshakes/ik.ts
mpetrunic: Depends, if flag if everything is always required to be valid you can just allow propagation and whoever is using handshake should handle error and not upgrade connection (should dorp connection probably)

src/handshakes/abstract-handshake.ts
morrigan: I don't have any strong opinion on this so...

mpetrunic: You ain't writing unit tests for handshake so it doesn't have much impact. (It's much more "testable" code if methods are not wrapped inside class). It could have impact if third handshake uses different stuff

morrigan: Got your point. Leaving for now as it's not the case

src/handshakes/ik.ts
morrigan: This is an interesting case. I have no idea if those validity flags are just Go thing or are required for the connection handling. However, execution of this part is different in the current JS version where if this line returns invalid, it doesn't continue execution which is the opposite of Go's.

src/handshakes/abstract-handshake.ts
```diff
@@ -1,13 +1,13 @@
...
-import { AEAD, x25519, HKDF, SHA256 } from 'bcrypto';
+import { AEAD, x25519, SHA256 } from 'bcrypto';
 
 import {bytes, bytes32, uint32} from "../@types/basic";
 import {CipherState, MessageBuffer, SymmetricState} from "../@types/handshake";
 import {getHkdf} from "../utils";
 
-export abstract class AbstractHandshake {
-  protected minNonce = 0;
+const minNonce = 0;
```
mpetrunic: constants are in uppercase and should probably be exporeted like:
`export const MIN_NONCE = 0;`

PR #12, opened 2020-01-07
XX fallback

morrigan:
- Added XX fallback in XX handshake
 - Test for XX fallback
 - Updated encoding to match Go implementation


morrigan:
@yusefnapora you can take a look as well if you can :)

src/handshake-ik.ts
```diff
@@ -0,0 +1,76 @@
...
+  private connection: WrappedConnection;
+  private remotePeer: PeerId;
+  private ik: IK;
+
+  constructor(
+    isInitiator: boolean,
+    libp2pPrivateKey: bytes,
+    libp2pPublicKey: bytes,
+    prologue: bytes32,
+    staticKeys: KeyPair,
```
mpetrunic: staticKeypair*

src/handshake-xx-fallback.ts
```diff
@@ -0,0 +1,89 @@
...
+  getHandshakePayload,
+  signEarlyDataPayload,
+  signPayload,
+  verifySignedPayload,
+} from "./utils";
+import { logger } from "./logger";
+import { WrappedConnection } from "./noise";
+import {decode0, decode1, encode1} from "./encoder";
+
+export class Handshake extends XXHandshake {
```
mpetrunic: XXFallbackHandshake extends XXHandshake*

src/handshake-xx-fallback.ts
```diff
@@ -0,0 +1,89 @@
...
+      });
+      logger("XX Fallback Stage 0 - Responder received first message.");
+    }
+  }
+
+  // stage 1
+  public async exchange(): Promise<void> {
+    if (this.isInitiator) {
+      logger('XX Fallback Stage 1 - Initiator waiting to receive first message from responder...');
+      const receivedMessageBuffer = decode1((await this.connection.readLP()).slice());
```
mpetrunic: why slice? It will create new buffer instead of using existing (I don't think connection will care if you modify that)

yusefnapora:
@morrigan if `fallback == true` in that `runhandshake_xx` method & we're the initiator, it means that we've sent a stage zero `IK` message and got a response that failed to decrypt as `IK`. So in `XXfallback`, we re-init the state to pretend we'd sent a stage zero `XX message instead, and then take the `IK` response we've already read from the stream and process it as the `XX` response we'd normally be reading in stage one of `XX`.

The reason we can't read the response message directly from the stream is just that it was already consumed by the `IK` handshake code, so it has to get passed in in the `initalMsg` variable.

morrigan:
> @morrigan if `fallback == true` in that `runhandshake_xx` method & we're the initiator, it means that we've sent a stage zero `IK` message and got a response that failed to decrypt as `IK`. So in `XXfallback`, we re-init the state to pretend we'd sent a stage zero `XX` message instead

Yes, I got this in initiator stage 0 (as the initiator is the IK initiator as well).

> and then take the `IK` response we've already read from the stream and process it as the `XX` response we'd normally be reading in stage one of `XX`.

Yes, responder stage 0. What about then stage 1 of initiator? How can the same message be processed as received by both parties?

yusefnapora:
I think that the way the go code is written, stage zero for the initiator ends once we've sent the initial `XX` handshake message. In `XXfallback`, we skip that step since the stage zero `IK` message served the same purpose. So `XXfallback` jumps to stage one of the regular `XX` flow, which is reading the first message from the responder.

I think the confusion is that the initiator and responder enter the "fallback" flow at different stages. For the responder, `XXfallback` gets triggered in stage zero, when processing the initiator's first message ("message A" in the terms used in this repo). In this case, `initialMsg` in the go code contains Message A sent by the initiator.

But the initiator doesn't fallback until their stage one, when processing message B (sent by the responder in the responder's stage zero). This starts out as stage one of the initiator's `IK` handshake, but when that fails, the initiator "skips ahead" to stage one of `XX`. In this case, `initialMsg` contains Message B, sent by the responder.

Does that make sense?

morrigan:
@yusefnapora yes, I think I got it! Thanks a lot for explaining it!

Ok, seems fine, I've just didn't mock things in the test well. Will fix that and continue with the integration of IK handshake flow, then test better all together :)

.eslintrc
```diff
@@ -15,6 +15,7 @@
     "@typescript-eslint/indent": ["error", 2],
     "@typescript-eslint/no-use-before-define": "off",
     "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/interface-name-prefix": "off",
```
mpetrunic: you could set it to force I prefix

src/noise.ts
```diff
@@ -76,29 +97,83 @@ export class Noise implements NoiseConnection {
...
+    // TODO: Implement noise pipes
+
+    if (false) {
+      let IKhandshake;
+      try {
+        IKhandshake = await this.performIKHandshake(params);
+        return IKhandshake;
+      } catch (e) {
+        // XX fallback
+        const ephemeralKeys = IKhandshake.getRemoteEphemeralKeys();
```
mpetrunic: but IKhandshake will be null in case of exception

morrigan: Not necessarily. Will see when IK is implemented.

mpetrunic: according to current code, in case of exception in `performIKHandshake` IKhandshake will be null in catch clause

morrigan: ? The whole code for that is TODO

morrigan: ? The whole code for that is under TODO

PR #13, opened 2020-01-07
Fix providing private key

### LINKED ISSUE #6: libp2p keys may be other types than ed25519
Hey 👋thanks for making this!

I noticed as I was reviewing this that the constructor accepts a `privateKey` argument which is used for signing / verifying payloads, and that the key is assumed to be an ed25519 key. Since libp2p keys can also be RSA or secp256k1 keys, this assumption might break.

Instead, you can grab the private key from the `localPeer` argument that's provided to `secureInbound` and `secureOutbound`. This will have a `sign` method you can use for signing the handshake payload. To verify the remote party's signatures, you can unmarshal the key and call its `verify` method, instead of calling `ed25519.verify`.

morrigan:
Resolves #6 

cc @yusefnapora

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,4 +1,4 @@
-import { x25519, ed25519, HKDF, SHA256 } from 'bcrypto';
+import { x25519, HKDF, SHA256 } from 'bcrypto';
 import protobuf from "protobufjs";
 import { Buffer } from "buffer";
 import PeerId from "peer-id";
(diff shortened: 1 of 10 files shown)


src/utils.ts
```diff
@@ -23,6 +23,22 @@ export function generateKeypair(): KeyPair {
...
+  localPeer: PeerId,
+  staticPublicKey: bytes,
+  earlyData?: bytes,
+): Promise<bytes> {
+  const signedPayload = await signPayload(localPeer, getHandshakePayload(staticPublicKey));
+  const signedEarlyDataPayload = await signEarlyDataPayload(localPeer, earlyData || Buffer.alloc(0));
+
+  return await createHandshakePayload(
+    localPeer.marshalPubKey(),
+    localPeer.marshalPrivKey(),
```
yusefnapora: It looks like you could remove the `libp2pPrivateKey` argument from the `createHandshakePayload` function, as it's not being used.

PR #18, opened 2020-01-13
Noise pipes

### LINKED ISSUE #14: Noise pipes
Implement Noise pipes flow

morrigan:
Resolves #14

### DIFF
--- a/src/keycache.ts
+++ b/src/keycache.ts
@@ -20,11 +20,11 @@ class Keycache {
     }
   }
 
-  public async load(peerId: PeerId): Promise<bytes32> {
+  public async load(peerId: PeerId): Promise<bytes32|null> {
     const release = await this.mutex.acquire();
     let key;
     try {
-      key = this.storage.get(peerId.id);
+      key = this.storage.get(peerId.id) || null;
     } finally {
       release();
     }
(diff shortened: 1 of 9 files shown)


src/keycache.ts
```diff
@@ -0,0 +1,46 @@
+import {Mutex} from 'async-mutex';
+import {PeerId} from "./@types/libp2p";
+import {bytes, bytes32} from "./@types/basic";
+
+/**
+ * Storage for static keys of previously connected peers.
+ */
+class Keycache {
+  private mutex: Mutex;
```
mpetrunic: why do you need mutex, storage should be thread safe by default

morrigan: Handshakes are async and I guess a lot of connections could go through at the same time. Those were my thoughts... Not sure what do you mean by default?

mpetrunic: It doesn't matter if they are async, as long as you are dealing with javascript object it will be thread safe (methods on storage will be executed sequentially). It makes sense using mutex if you are dealing with native database api or maybe using worker threads

mpetrunic: https://stackoverflow.com/questions/7238586/do-i-need-to-be-concerned-with-race-conditions-with-asynchronous-javascript/7238663#7238663

morrigan: The answer is pretty vague, especially the second part, but it could be summed up as "only if using some native code". I guess I've been reading the wrong things.

morrigan:
Forgot @yusefnapora, if you want to take a look :)

PR #20, opened 2020-01-29
Set prologue to empty buffer

### LINKED ISSUE #8: Noise prologue is set to the protocol id string
I noticed that the noise prologue is initialized with the ID string (`/noise`), which will cause interop issues with the go implementation, which doesn't set the prologue at all.

Ideally, both implementations would set the prologue to whatever bytes were actually exchanged over the wire prior to the noise handshake occurring. This would be the back-and-forth from multistream-select that selects Noise as the crypto channel. However, I don't think we have a good interface for providing that data in either go or js 😞 

In the long term, we're planning to move away from negotiating crypto channels entirely. Instead, the crypto protocol will be part of the multiaddr. So there won't be a need for a prologue in that case.

We could potentially update the secure transport interfaces in js and go to accept optional prologue data, but we haven't scoped that work out yet, and given that we want to move away from negotiation it may not be worth the effort. 

In the meantime, it seems best to just ditch the prologue entirely, rather than set it to a fixed string.

morrigan:
Resolves #8

### DIFF
--- a/src/noise.ts
+++ b/src/noise.ts
@@ -31,7 +31,7 @@ type HandshakeParams = {
 export class Noise implements INoiseConnection {
   public protocol = "/noise";
 
-  private readonly prologue = Buffer.from(this.protocol);
+  private readonly prologue = Buffer.alloc(0);
   private readonly staticKeys: KeyPair;
   private readonly earlyData?: bytes;
   private useNoisePipes: boolean;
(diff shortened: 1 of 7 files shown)


src/noise.ts
```diff
@@ -31,7 +31,7 @@ type HandshakeParams = {
 export class Noise implements INoiseConnection {
   public protocol = "/noise";
 
-  private readonly prologue = Buffer.from(this.protocol);
+  private readonly prologue = Buffer.alloc(0);
```
mpetrunic: is there some requirement for prologue size?

morrigan: It's not in the spec but according to Go implementation - no. https://github.com/libp2p/go-libp2p-noise/blob/master/protocol.go#L95

PR #32, opened 2020-02-07
Fix IK flow

mpetrunic:
This is alternative to #31 

- remote peer is deducted from payload in message respective to the handshake type
- ik is tried if 
   - initiator + noise pipes + has remote static key
   - responder + noise pipes

Some tests are failing:
   - 2 are wrong as they assume you will receive remote peer on inbound connection
   - rest are failing because "Ephemeral keys doesn't exists"

### DIFF
--- a/src/noise.ts
+++ b/src/noise.ts
@@ -181,8 +181,8 @@ export class Noise implements INoiseConnection {
       await handshake.exchange();
       await handshake.finish();
 
-      if (this.useNoisePipes && remotePeer) {
-        KeyCache.store(remotePeer, handshake.getRemoteStaticKey());
+      if (this.useNoisePipes && handshake.remotePeer) {
+        KeyCache.store(handshake.remotePeer, handshake.getRemoteStaticKey());
       }
     } catch (e) {
       throw new Error(`Error occurred during XX handshake: ${e.message}`);
--- a/src/noise.ts
+++ b/src/noise.ts
@@ -1,20 +1,20 @@
-import { x25519 } from 'bcrypto';
-import { Buffer } from "buffer";
+import {x25519} from 'bcrypto';
+import {Buffer} from "buffer";
 import Wrap from 'it-pb-rpc';
 import DuplexPair from 'it-pair/duplex';
 import ensureBuffer from 'it-buffer';
 import pipe from 'it-pipe';
 import lp from 'it-length-prefixed';
 
-import { XXHandshake } from "./handshake-xx";
-import { IKHandshake } from "./handshake-ik";
-import { XXFallbackHandshake } from "./handshake-xx-fallback";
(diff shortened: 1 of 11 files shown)


src/noise.ts
```diff
@@ -111,26 +111,38 @@ export class Noise implements INoiseConnection {
    */
   private async performHandshake(params: HandshakeParams): Promise<IHandshake> {
     const payload = await getPayload(params.localPeer, this.staticKeys.publicKey, this.earlyData);
-
-    const remoteStaticKey = KeyCache.load(params.remotePeer);
+    let tryIK = this.useNoisePipes;
+    if(params.isInitiator && KeyCache.load(params.remotePeer) === null) {
```
morrigan: What should happen if responder doesn't have the remote static key in cache? One test fails because of this condition.

mpetrunic: responder doesn't need to know initiator remote static key, he decrypts with his private static key as message is encrypted with his public static key. Spec:

mpetrunic:
```
Later, Alice can open a new Noise connection to Bob using the IK pattern. This is a zero-RTT handshake that uses the cached static key to encrypt the initial handshake message.

If Alice attempts an IK handshake but Bob has changed his static Noise key, Bob will fail to decrypt the handshake message.
```

morrigan: Oh, yeah, indeed. Test is wrong :)

src/utils.ts
```diff
@@ -63,14 +63,37 @@ export async function signPayload(peerId: PeerId, payload: bytes): Promise<bytes
   return peerId.privKey.sign(payload);
 }
 
+export async function getPeerIdFromPayload(payload: bytes): Promise<PeerId> {
+  const decodedPayload = await decodePayload(payload);
+  return await PeerId.createFromPubKey(Buffer.from(decodedPayload.identityKey));
+}
+
+async function decodePayload(payload: bytes){
```
mpetrunic: add return type

src/utils.ts
```diff
@@ -63,14 +63,37 @@ export async function signPayload(peerId: PeerId, payload: bytes): Promise<bytes
...
+ * @param {bytes} noiseStaticKey - owner's noise static key
+ * @param {bytes} plaintext - encoded payload
+ * @param {PeerId} remotePeer - (optional) owner's libp2p peer ID
+ * @returns {Promise<PeerId>} - peer ID of payload owner
+ */
+export async function verifySignedPayload(
+  noiseStaticKey: bytes,
+  plaintext: bytes,
+  remotePeer?: PeerId
+): Promise<PeerId> {
```
mpetrunic: It doesn't make sense for this method to return PeerId, it's called verifySignedPayload it can only return boolean or void.

It makes sense for this method to receive decodedPayload instead of plaintext

PR #40, opened 2020-03-03
remove empty ephemeral key from XX message C

yusefnapora:
This branch makes the final message in the XX sequence conform to the Noise framework spec by removing the empty ephemeral key, as described in https://github.com/libp2p/go-libp2p-noise/issues/54.

I've merged in the fix on the go side, so now interop tests are failing against master. I have tested the new Go master branch against this js branch though, and it works.

Thanks again @sinkingsugar for filing the issue!


mpetrunic:
Seems good, I'll check tests locally

PR #47, opened 2020-04-03
Fix bcrypto imports

mpetrunic:
Pros:
- reduces bcrypto size in bundle by ~50%

Cons:
- slower in nodejs (it's just pure js implementations)

Current bundle size iz 913Kb (214kb gzipped). node-forge and bcrypto are still large
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Bugfixes
+- changed bcrypto imports to use pure js versions (web bundle size reduction)
+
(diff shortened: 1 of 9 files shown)


morrigan:
Would the best thing be to get rid of bcrypto? As I can see, we have now alternative options for each import we have right now.

mpetrunic:
Wha

> Would the best thing be to get rid of bcrypto? As I can see, we have now alternative options for each import we have right now.

What would be alternative for hkdf and x25519?

morrigan:
> Wha
> 
> > Would the best thing be to get rid of bcrypto? As I can see, we have now alternative options for each import we have right now.
> 
> What would be alternative for hkdf and x25519?

https://www.npmjs.com/package/js-crypto-hkdf ?
x25519 has multiple options as it is widely used I guess

Issue #27, opened 2020-02-07
Webpack bundle too large


mpetrunic:
- bcrypto is using "import" instead of require which pull entire bcrypto lib and adds ~1.2 Mb to webpack bundle

mpetrunic:
https://bundlephobia.com/result?p=libp2p-noise@1.0.0-rc.9

morrigan:
Reduced to 681kB - https://bundlephobia.com/result?p=libp2p-noise@1.1.0
Not great, not terrible.

Closing in favor that libp2p-crypto will it's reduce size.

jacobheun:
Just FYI, I am working on downstream integration of this in js-ipfs, https://github.com/ipfs/js-ipfs/pull/3004. The bundlesize change there is only around 73KB. Not great, but not as much as I was expecting. I'll be working on this over the next few weeks and will submit any updates here as I find them.

> libp2p-crypto will it's reduce size

I am planning on making libp2p-crypto composable to make reducing the size easier, but likely won't get to that until around July.

mpetrunic:
Would be nice if you added x25519, aead(chacha-poly) and hkdf support, we tried swapping those but none of the x25519 alternatives worked in interop.

PR #51, opened 2020-04-06
Replace bcrypto with standalone libraries

mpetrun5:
Removes bcrypto from dependencies.

Replaced with: 
 - [futoin-hdkf](https://www.npmjs.com/package/futoin-hkdf)
 - [aead-js](https://www.npmjs.com/package/aead-js)
 - [tweetnacl](https://www.npmjs.com/package/tweetnacl)

Related to issue: #27 

cc: @mpetrunic

### DIFF
--- a/src/handshakes/abstract-handshake.ts
+++ b/src/handshakes/abstract-handshake.ts
@@ -106,7 +106,7 @@ export abstract class AbstractHandshake {
 
   protected dh(privateKey: bytes32, publicKey: bytes32): bytes32 {
     try {
-      return Buffer.from(box.before(publicKey, privateKey))
+      return Buffer.from(box.before(publicKey, privateKey));
     } catch (e) {
       logger(e.message);
       return Buffer.alloc(32);
@@ -120,7 +120,7 @@ export abstract class AbstractHandshake {
   protected getHash(a: bytes, b: bytes): bytes32 {
     return Buffer.from(
       hash.sha256().update(Buffer.from([...a, ...b])).digest('hex')
-    )
+    );
   }
 
   protected mixKey(ss: SymmetricState, ikm: bytes32): void {
--- a/src/handshakes/abstract-handshake.ts
+++ b/src/handshakes/abstract-handshake.ts
@@ -107,7 +107,7 @@ export abstract class AbstractHandshake {
   protected dh(privateKey: bytes32, publicKey: bytes32): bytes32 {
     try {
       const sharedKey = box.before(publicKey, privateKey)
-      return Buffer.from(sharedKey);
+      return Buffer.from(sharedKey.buffer, sharedKey.byteOffset, sharedKey.length);
     } catch (e) {
       logger(e.message);
       return Buffer.alloc(32);
--- a/src/handshakes/abstract-handshake.ts
+++ b/src/handshakes/abstract-handshake.ts
@@ -106,7 +106,8 @@ export abstract class AbstractHandshake {
 
   protected dh(privateKey: bytes32, publicKey: bytes32): bytes32 {
     try {
-      return Buffer.from(box.before(publicKey, privateKey), 0, 32);
+      const sharedKey = box.before(publicKey, privateKey)
+      return Buffer.from(sharedKey);
     } catch (e) {
       logger(e.message);
       return Buffer.alloc(32);
@@ -118,9 +119,8 @@ export abstract class AbstractHandshake {
   }
 
   protected getHash(a: bytes, b: bytes): bytes32 {
-    return Buffer.from(
-      hash.sha256().update(Buffer.from([...a, ...b])).digest('hex')
-    );
+    const hashValue = hash.sha256().update(Buffer.from([...a, ...b])).digest();
+    return Buffer.from(hashValue);
   }
 
   protected mixKey(ss: SymmetricState, ikm: bytes32): void {
--- a/src/handshakes/abstract-handshake.ts
+++ b/src/handshakes/abstract-handshake.ts
@@ -1,5 +1,7 @@
 import {Buffer} from "buffer";
-import { AEAD, x25519, SHA256 } from 'bcrypto';
+import hash from 'hash.js';
+import {box} from 'tweetnacl';
+import {AEAD} from 'aead-js';
 
 import {bytes, bytes32, uint32} from "../@types/basic";
 import {CipherState, MessageBuffer, SymmetricState} from "../@types/handshake";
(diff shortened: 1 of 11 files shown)


src/handshakes/abstract-handshake.ts
```diff
@@ -104,10 +106,7 @@ export abstract class AbstractHandshake {
 
   protected dh(privateKey: bytes32, publicKey: bytes32): bytes32 {
     try {
-      const derived = x25519.derive(publicKey, privateKey);
-      const result = Buffer.alloc(32);
-      derived.copy(result);
-      return result;
+      return Buffer.from(box.before(publicKey, privateKey), 0, 32);
```
mpetrunic, line 109: What are you trying to achieve here?
Buffer view over 0...32 byte of output?

src/handshakes/abstract-handshake.ts
```diff
@@ -119,7 +118,9 @@ export abstract class AbstractHandshake {
   }
 
   protected getHash(a: bytes, b: bytes): bytes32 {
-    return SHA256.digest(Buffer.from([...a, ...b]));
+    return Buffer.from(
+      hash.sha256().update(Buffer.from([...a, ...b])).digest('hex')
```
mpetrunic, line 122: This hash is invalid. Buffer.from uses utf-8 encoding by default.

To reduce memory allocation:
```javascript
const hash = hash.sha256().update(Buffer.from([...a, ...b])).digest();
Buffer.from(hash.buffer, hash.offset, hash.length);
```
https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length

src/utils.ts
```diff
@@ -111,5 +111,9 @@ export function getHkdf(ck: bytes32, ikm: bytes): Hkdf {
 }
 
 export function isValidPublicKey(pk: bytes): boolean {
-  return x25519.publicKeyVerify(pk.slice(0, 32));
+  if(pk.length !== 32 || pk.equals(Buffer.alloc(32))){
+    return false;
+  }
+
+  return true;
```
mpetrunic, line 118: @morrigan Any idea how to implement proper public key verification?

test/noise.test.ts
```diff
@@ -3,8 +3,8 @@ import DuplexPair from 'it-pair/duplex';
 import {Noise} from "../src";
 import {createPeerIdsFromFixtures} from "./fixtures/peer";
 import Wrap from "it-pb-rpc";
-import {random} from "bcrypto";
 import sinon from "sinon";
+import crypto from 'crypto';
```
mpetrunic, line 7: You didn't list this dependency.
Also you have this method in libp2p-crypto
https://github.com/libp2p/js-libp2p-crypto#cryptorandomBytesnumber

src/utils.ts
```diff
@@ -10,8 +11,9 @@ import {pb} from "./proto/payload";
 const NoiseHandshakePayloadProto = pb.NoiseHandshakePayload;
 
 export function generateKeypair(): KeyPair {
-  const privateKey = x25519.privateKeyGenerate();
-  const publicKey = x25519.publicKeyCreate(privateKey);
+  const keyPair = box.keyPair();
+  const publicKey = Buffer.from(keyPair['publicKey']);
+  const privateKey = Buffer.from(keyPair['secretKey']);
```
mpetrunic, line 16: any reason why you didn't do `keyPair.publicKey`?

src/handshakes/abstract-handshake.ts
mpetrun5: I unnecessarily tried to replicate the old logic.

src/handshakes/abstract-handshake.ts
```diff
@@ -106,7 +106,8 @@ export abstract class AbstractHandshake {
 
   protected dh(privateKey: bytes32, publicKey: bytes32): bytes32 {
     try {
-      return Buffer.from(box.before(publicKey, privateKey), 0, 32);
+      const sharedKey = box.before(publicKey, privateKey)
+      return Buffer.from(sharedKey);
```
mpetrunic, line 110:
```suggestion
      return Buffer.from(sharedKey.buffer, sharedKey.offset, sharedKey.length);
```
It won't allocate new array

src/handshakes/abstract-handshake.ts
```diff
@@ -118,9 +119,8 @@ export abstract class AbstractHandshake {
   }
 
   protected getHash(a: bytes, b: bytes): bytes32 {
-    return Buffer.from(
-      hash.sha256().update(Buffer.from([...a, ...b])).digest('hex')
-    );
+    const hashValue = hash.sha256().update(Buffer.from([...a, ...b])).digest();
+    return Buffer.from(hashValue);
```
mpetrunic, line 123: same as above

mpetrun5: hashValue is number[]. it doesn't have buffer or offset.

mpetrunic: Could you try it. It's probably Uint8Array but it's displayed as number[] in logs
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array

mpetrunic: nvm, it's regular Array, wierd. I guess it's fine like this

src/utils.ts
morrigan: Can't find anything. I guess no big deal as it won't pass DH otherwise

morrigan:
What's the size now?

mpetrun5:
image
Webpack analyzer is kinda broken in this image, but It says 1020 kib.
https://bundlephobia.com/scan-results?packages=@types/bn.js@4.11.6,aead-js@0.1.0,bn.js@5.0.0,buffer@5.4.3,debug@4.1.1,futoin-hkdf@1.3.2,hash.js@1.1.7,it-buffer@0.1.1,it-length-prefixed@3.0.0,it-pair@1.0.0,it-pb-rpc@0.1.8,it-pipe@1.1.0,libp2p-crypto@0.17.1,peer-id@0.13.5,protobufjs@6.8.8,tweetnacl@1.0.3

It is not much better.

mpetrunic:
I wonder if we can reduce amount of stuff imported from libp2p-crypto since that is pulling node-forge. I guess that something guys from libp2p will do

morrigan:
Surprisingly, it's not even better than #47 :/

test/noise.test.ts
```diff
@@ -3,8 +3,8 @@ import DuplexPair from 'it-pair/duplex';
 import {Noise} from "../src";
 import {createPeerIdsFromFixtures} from "./fixtures/peer";
 import Wrap from "it-pb-rpc";
-import {random} from "bcrypto";
 import sinon from "sinon";
+import crypto from 'libp2p-crypto';
```
mpetrunic, line 7: try to
`import {randomBytes} from "libp2p-crypto"`

mpetrun5:
It shows no difference in webpack-bundle-analyzer.

mpetrunic:
@morrigan Can you re-review it, it somewhat smaller now, but I don't think it can get better without fixing js-libp2p-crypto. That said not sure if we should just apply last changes in #47 or should we merge this?

mpetrunic:
I will make a release out of it and try it in interop repo with tests against go version

Issue #16, opened 2020-01-10
Dump session keys for debugging


morrigan:
[https://github.com/libp2p/go-libp2p-noise/issues/24](https://github.com/libp2p/go-libp2p-noise/issues/24)

morrigan:
@mpetrunic @mpetrun5 I think it's ok for debugging that we display both static and ephemeral keys but also the encryption key we use after the handshake is completed. The first two might be useful for interops but the latter is probably useful for libp2p communications debugging.

PR #52, opened 2020-04-10
Dump session keys

mpetrun5:
Log session keys when created and exchanged.

Related: #16


src/constants.ts
```diff
@@ -1,3 +1,6 @@
 export const NOISE_MSG_MAX_LENGTH_BYTES = 65535;
 export const NOISE_MSG_MAX_LENGTH_BYTES_WITHOUT_TAG = NOISE_MSG_MAX_LENGTH_BYTES - 16;
 
+export const DUMP_SESSION_KEYS = false;
```
mpetrunic, line 4:
```suggestion
export const DUMP_SESSION_KEYS = process.env.NOISE_DUMP_KEYS;
```

mpetrun5: You already need DEBUG=libp2p:session as an enviroment variable to show session keys. This  const made sense so you won't accidentally leave an enviroment variable active and show them. So in my opinion either we don't need this variable or leave as a constant.

mpetrunic: oh don't do that, libp2p:session is meant for logs called js-libp2p-session. Usually people will log libp2p stuff with DEBUG=libp2p:* which would log keys by accident and we should avoid that

mpetrun5: Definitely should change DEBUG to something more specific to this, eg. noise:session-keys?

mpetrunic: it should stay libp2p:noise, but logging of keys should depend on additional env variable

src/logger.ts
```diff
@@ -1,2 +1,51 @@
 import debug from "debug";
-export const logger = debug('libp2p:noise');
+import {DUMP_SESSION_KEYS} from './constants';
+import { KeyPair } from "./@types/libp2p";
+import { NoiseSession, SymmetricState } from "./@types/handshake";
+
+let keyLogger;
+if(DUMP_SESSION_KEYS){
+  keyLogger = debug('libp2p:session')
```
mpetrunic, line 8: Use existing logger instead of debug('libp2p:session')

src/handshake-xx.ts
```diff
@@ -50,11 +58,13 @@ export class XXHandshake implements IHandshake {
...
   public async propose(): Promise<void> {
+    logLocalStaticKeys(this.session.hs.s)
     if (this.isInitiator) {
       logger("Stage 0 - Initiator starting to send first message.");
       const messageBuffer = this.xx.sendMessage(this.session, Buffer.alloc(0));
       this.connection.writeLP(encode0(messageBuffer));
       logger("Stage 0 - Initiator finished sending first message.");
+      logLocalEphemeralKeys(this.session.hs.e)
     } else {
       logger("Stage 0 - Responder waiting to receive first message...");
```
morrigan, line 69: I think you are missing `logLocalStaticKeys` for the responder too

PR #53, opened 2020-04-17
Early data api

mpetrun5:
Set remote early data when remote peer payload is verified and return from secureOutbound and secureInbound.

Related to issue: #53


src/handshake-xx.ts
```diff
@@ -146,4 +150,10 @@ export class XXHandshake implements IHandshake {
       return encryption ? session.cs2 : session.cs1;
     }
   }
+
+  protected setRemoteEarlyData(data: Uint8Array|null|undefined): void {
+    if(data){
+      this.remoteEarlyData = Buffer.from(data.buffer, data.byteOffset, data.length);
```
morrigan, line 156: I think this is same as`Buffer.from(data)` ?

mpetrunic: Yours will allocate new array, while this is just view over existing array:
https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length

PR #56, opened 2020-04-22
Revert "Replace bcrypto with standalone libraries"

morrigan:
Reverts NodeFactoryIo/js-libp2p-noise#51

Screenshot 2020-04-22 at 19 30 04


package.json
```diff
@@ -67,22 +67,16 @@
...
     "debug": "^4.1.1",
-    "futoin-hkdf": "^1.3.2",
-    "hash.js": "^1.1.7",
     "it-buffer": "^0.1.1",
     "it-length-prefixed": "^3.0.0",
     "it-pair": "^1.0.0",
     "it-pb-rpc": "^0.1.8",
     "it-pipe": "^1.1.0",
-    "libp2p-crypto": "^0.17.6",
+    "libp2p-crypto": "^0.17.1",
```
mpetrunic, line 78: why not latest? available

package.json
```diff
@@ -67,22 +67,16 @@
...
     "peer-id": "^0.13.5",
-    "protobufjs": "6.8.8",
-    "tweetnacl": "^1.0.1"
-  },
-  "resolutions": {
-    "bn.js": "4.4.0"
```
mpetrunic, line 86: This reduces bundle size because it forces all deps to use same bn version

test/noise.test.ts
```diff
@@ -3,8 +3,8 @@ import DuplexPair from 'it-pair/duplex';
 import {Noise} from "../src";
 import {createPeerIdsFromFixtures} from "./fixtures/peer";
 import Wrap from "it-pb-rpc";
+import {random} from "bcrypto";
```
mpetrunic, line 6: It's better to use libp2p-crypto random. This will pull whole bcrypto again

package.json
morrigan: idk, this was not on purpose. I'll update

PR #64, opened 2020-06-18
feat: use aegir

hugomrdias:
needs:
https://github.com/ipfs/aegir/pull/584


package.json
```diff
@@ -18,53 +18,23 @@
...
-    "bundle": "webpack --config webpack.bundle.config.js",
-    "build:node": "babel --no-babelrc --config-file ./babel.config.json src --copy-files -x .ts -d dist --source-maps",
-    "build:web": "babel --no-babelrc --config-file ./babel.web.config.json src --copy-files -x .ts -d lib --source-maps",
-    "build:types": "tsc --declaration --outDir dist --emitDeclarationOnly",
-    "proto:gen": "pbjs -t static-module -o ./src/proto/payload.js ./src/proto/payload.proto && pbts -o ./src/proto/payload.d.ts ./src/proto/payload.js && yarn run lint --fix",
-    "check-types": "tsc --noEmit",
-    "lint": "eslint --ext .ts src/",
+    "build": "aegir build --ts",
+    "lint": "aegir lint --ts",
     "pretest": "yarn check-types",
```
mpetrunic, line 23: since build is ran using tsc (it will check types also), we don't need this to do type checking here

package.json
```diff
@@ -18,53 +18,23 @@
...
-    "test": "yarn run test:node && yarn run test:web",
-    "test:node": "mocha -r ./babel-register.js \"test/**/*.test.ts\"",
-    "test:web": "karma start"
+    "test": "aegir test --ts",
+    "test:node": "aegir test -t node --ts",
+    "test:browser": "aegir test -t browser --ts",
+    "check-types": "tsc --noEmit",
+    "proto:gen": "pbjs -t static-module -o ./src/proto/payload.js ./src/proto/payload.proto && pbts -o ./src/proto/payload.d.ts ./src/proto/payload.js && yarn run lint --fix"
```
mpetrunic, line 28: I don't think this would work in packaged library. tsc doesn't copy non ts files. We could either enable allowJs in tsconfig.json or add npm script to copy it manually in postbuild task

hugomrdias:
Where's the module prop ? i cant see it

mpetrunic:
> Where's the module prop ? i cant see it

You already removed it in last commit :smile:

package.json
```diff
@@ -1,13 +1,9 @@
 {
   "name": "libp2p-noise",
   "version": "1.1.2",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
```
mpetrunic, line 5: We should leave this, or typescript users won't detect typings

package.json
```diff
@@ -1,13 +1,9 @@
...
-  "module": "lib/index.js",
+  "main": "dist/src/index.js",
   "files": [
-    "dist",
-    "lib",
-    "src"
```
mpetrunic, line 10: I would leave src as we need to provide sourceMaps so we can have info where error occured in typescript file rather than builded js file

hugomrdias:
did i? nice :)
I would hand this over to you now if its ok, theres a lot of linter fixes that im probably not the best person to fix and i will fix any issues on aegir side, sounds good ?

mpetrunic:
> did i? nice :)
> I would hand this over to you now if its ok, theres a lot of linter fixes that im probably not the best person to fix and i will fix any issues on aegir side, sounds good ?

Yeah sure. Does `aegir lint` support --fix argument?

hugomrdias:
> Yeah sure. Does `aegir lint` support --fix argument?

yes

hugomrdias:
Plus if you run --help you will see how to forward args to the underlying cli. Not all cmds support this but most do.

package.json
```diff
@@ -18,53 +14,21 @@
...
-    "@babel/preset-typescript": "^7.6.0",
-    "@babel/register": "^7.6.2",
-    "@babel/runtime": "^7.6.3",
     "@types/bl": "^2.1.0",
     "@types/chai": "^4.2.4",
     "@types/mocha": "^5.2.7",
-    "@typescript-eslint/eslint-plugin": "^2.6.0",
-    "@typescript-eslint/parser": "^2.6.0",
-    "babel-loader": "^8.1.0",
+    "aegir": "ipfs/aegir#feat/cosmiconfig",
```
vasco-santos, line 28: Ups, This got merged without aegir updated

mpetrunic: Im gonna open few pr with some nitpicks and lint fixes so I will update before next release 🙂

PR #67, opened 2020-06-19
Fix lint errors and tests

mpetrunic:
- enabled source maps
- fixed some configs
- fixed lint errors
- node tests passing

TODO:
- [ ] web tests passing


mpetrunic:
Seems like bcrypto library relies on Buffer polyfill in browser, any idea how to fix that @hugomrdias

hugomrdias:
> Seems like bcrypto library relies on Buffer polyfill in browser, any idea how to fix that @hugomrdias

You need to run aegir cmds with --node true, we removed the dependency on node globals and builtins in all the other ipfs/libp2p repos and deps https://github.com/ipfs/js-ipfs/issues/2924 . This was because the next version of webpack (v5) is going to remove automatic injection of node stuff. 
So you should probably do the same for bcrypto, because this flag is going way soon.

package.json
```diff
@@ -2,8 +2,10 @@
   "name": "libp2p-noise",
   "version": "1.1.2",
   "main": "dist/src/index.js",
+  "types": "dist/src/dist/index.d.ts",
```
hugomrdias, line 5: This seems to be a typo, but you really dont need to specify the types field when you have the .d.ts next to the source file

package.json
```diff
@@ -2,8 +2,10 @@
...
   "files": [
-    "dist"
+    "dist",
+    "src"
```
hugomrdias, line 8: why do you include the src ? its just taking space in node_modules, a consumer should only need the js source files and .d.ts.

package.json
```diff
@@ -27,6 +30,7 @@
     "@types/mocha": "^5.2.7",
     "aegir": "ipfs/aegir#feat/cosmiconfig",
     "chai": "^4.2.0",
+    "karma-mocha-webworker": "^1.3.0",
```
hugomrdias, line 33: You should not need this, karma has a weird issue with this module and fails to load it if the package doesnt come from npm. So when we release aegir properly this can be removed.

package.json
mpetrunic: for source maps links. Its common for typescript projects to provide source files for people that debug code.

package.json
mpetrunic: oh, did not know that, failed locally so ive added it. Will remove once we update aegir version

hugomrdias:
aegir 25.0.0 released https://github.com/ipfs/aegir/blob/master/CHANGELOG.md#2500-2020-06-20

package.json
mpetrunic: Even with released aegir it still says that it's missing

PR #70, opened 2020-08-11
chore: update deps to latest versions

achingbrain:
I've tried to make the minimum amount of changes necessary for this, since the underlying crypto libraries only support node Buffers or BufferLists there doesn't seem a lot of point in doing lots of conversions between Uint8Arrays and Buffers.

BREAKING CHANGES:

- All deps use Uint8Arrays in place of node Buffers


achingbrain:
cc @jacobheun @vasco-santos

package-lock.json
```diff
@@ -0,0 +1,19940 @@
+{
```
mpetrunic, line 1: We use yarn, would you mind deleting this file and updating yarn.lock

achingbrain: Done, my apologies!

mpetrunic:
@achingbrain Did you intentionally leave Buffers in noise constructor?

achingbrain:
> Did you intentionally leave Buffers in noise constructor?

Yes - the underlying `bcrypto` module is full of `Buffer.isBuffer` so you end up converting any `Uint8Array`s to `Buffer`s before passing them through.

I did original try to remove `Buffer`s entirely but it was a much more disruptive changeset and seemed like the wrong approach given you end up converting everything anyway.

achingbrain:
@mpetrunic could this be released please?

vasco-santos:
A note regarding releasing this. Please do not release it with `latest` tag.

While we do not have everything bubbled up to `js-libp2p` and `js-libp2p` released, we usually handle the breaking change releases by releasing it and set a `beta` tag for the release, keeping the latest as it was. We end up putting everything in `latest` at the same time to avoid people installing noise with this PR in `js-libp2p` until it is ready

achingbrain:
@mpetrunic from the commit log it looks like you're gearing up to release this as `1.2.0` - this should really go out as `2.0.0` as the dep updates will cause multiple versions of the peer-id module to appear in people's dep trees, some of which require node Buffers and some of which require Uint8Arrays which may break things.

mpetrunic:
Oh, was not aware of that, tnx. i dont think release pipepline supports beta tags so I will have to release it manually anyways

mpetrunic:
https://www.npmjs.com/package/libp2p-noise/v/2.0.0-beta released with beta tag

achingbrain:
I think @vasco-santos meant to release it with a beta npm dist tag, rather than putting it in the version number, that way we don’t need to do follow-up PRs to remove the extra specifier.

https://docs.npmjs.com/cli/dist-tag

mpetrunic:
Yes, there is also beta tag so you can use libp2p-noise@beta

achingbrain:
Ah, no, I meant 2.0.0 as beta, not 2.0.0-beta as beta, otherwise we have to re-PR to remove the -beta part of the version number. See how libp2p-tcp has done it, for example.

vasco-santos:
Yeah, I meant as @achingbrain said. Sorry for not being specific enough. 
We can update it by then now.

mpetrunic:
we have 2.0.0 now tagged with beta :slightly_smiling_face:

achingbrain:
Perfect, thank you!

PR #98, opened 2021-05-27
Replace bcrypto

### LINKED ISSUE #93: Replace bcrypto dep
We already tried this once and it didn't work but it was some time ago and some alternatives appeared.

Our bcrypto algo dependencies and proposed alternatives:
- x25519 
   - https://www.npmjs.com/package/ecies-25519
- aead (chacha-poly)
   -  https://www.npmjs.com/package/@stablelib/chacha20poly1305 (https://github.com/StableLib/stablelib/blob/master/packages/xchacha20poly1305/xchacha20poly1305.ts)
   - https://www.npmjs.com/package/chacha-js
- hkdf
   - https://www.npmjs.com/package/futoin-hkdf
   - https://www.npmjs.com/package/js-crypto-hkdf
   - https://www.npmjs.com/package/@stablelib/hkdf

mpetrun5:
Replaced:
 - bcrypto/x25519 with  https://github.com/StableLib/stablelib/tree/master/packages/x25519
 - bcrypto/aead with https://github.com/NodeFactoryIo/aead-js
 - bcrypto/hkdf with https://github.com/StableLib/stablelib/tree/master/packages/hkdf

resolves #93

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "@stablelib/sha256": "^1.0.1",
     "@stablelib/x25519": "^1.0.1",
     "aead-js": "^0.1.0",
-    "bcrypto": "^5.4.0",
     "debug": "^4.3.1",
     "it-buffer": "^0.1.1",
     "it-length-prefixed": "^5.0.2",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "util": false
   },
   "devDependencies": {
+    "@types/elliptic": "^6.4.12",
     "aegir": "^33.1.0",
     "benchmark": "^2.1.4",
     "buffer": "^6.0.3",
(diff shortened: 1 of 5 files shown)


mpetrun5:
> Why our non audited aeadjs lib instead of stablelib provided in issue?

For some reason stablelib has hardcoded 24bit nonce length and existing bcrypto implementation uses 12 bit nonce and I couldn't get it to work to be compatible. Our implementation is basically just a copy of bcrypto so it was a lot easier.

mpetrunic:
> > Why our non audited aeadjs lib instead of stablelib provided in issue?
> 
> For some reason stablelib has hardcoded 24bit nonce length and existing bcrypto implementation uses 12 bit nonce and I couldn't get it to work to be compatible. Our implementation is basically just a copy of bcrypto so it was a lot easier.

You were using wrong version xchacha20 uses 24byte nonce while chacha20 uses 12byte nonce. Stable lib has both, this is 12byte version: https://github.com/StableLib/stablelib/tree/master/packages/chacha20poly1305

PR #100, opened 2021-05-31
Remove bcypto

### LINKED ISSUE #93: Replace bcrypto dep

mpetrun5:
Replaced:
 - bcrypto/x25519 with  https://github.com/StableLib/stablelib/tree/master/packages/x25519
 - bcrypto/aead with https://github.com/StableLib/stablelib/tree/master/packages/chacha20poly1305
 - bcrypto/hkdf with https://github.com/StableLib/stablelib/tree/master/packages/hkdf

resolves #93

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "util": false
   },
   "devDependencies": {
-    "@types/elliptic": "^6.4.12",
     "aegir": "^33.1.0",
     "benchmark": "^2.1.4",
     "buffer": "^6.0.3",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "@stablelib/sha256": "^1.0.1",
     "@stablelib/x25519": "^1.0.1",
     "aead-js": "^0.1.0",
-    "bcrypto": "^5.4.0",
     "debug": "^4.3.1",
     "it-buffer": "^0.1.1",
     "it-length-prefixed": "^5.0.2",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "util": false
   },
   "devDependencies": {
+    "@types/elliptic": "^6.4.12",
     "aegir": "^33.1.0",
     "benchmark": "^2.1.4",
     "buffer": "^6.0.3",
@@ -46,6 +47,7 @@
     "sinon": "^10.0.0"
   },
   "dependencies": {
+    "@stablelib/x25519": "^1.0.1",
     "bcrypto": "^5.4.0",
     "debug": "^4.3.1",
     "it-buffer": "^0.1.1",
--- a/package.json
+++ b/package.json
@@ -47,10 +47,10 @@
     "sinon": "^10.0.0"
   },
   "dependencies": {
+    "@stablelib/chacha20poly1305": "^1.0.1",
     "@stablelib/hkdf": "^1.0.1",
     "@stablelib/sha256": "^1.0.1",
     "@stablelib/x25519": "^1.0.1",
-    "aead-js": "^0.1.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "util": false
   },
   "devDependencies": {
+    "@types/elliptic": "^6.4.12",
```
mpetrunic, line 39: I don't think you really need this.

src/handshakes/abstract-handshake.ts
```diff
@@ -80,15 +75,27 @@ export abstract class AbstractHandshake {
...
-    ciphertext = ciphertext.slice(0, ciphertext.length - 16)
-    ctx.init(k, nonce)
-    ctx.aad(ad)
-    ctx.decrypt(ciphertext)
-    // Decryption is done on the sent reference
-    return { plaintext: ciphertext, valid: ctx.verify(tag) }
+    const ctx = new ChaCha20Poly1305(k)
+    const encryptedMessage = ctx.open(
+      nonce,
+      new Uint8Array(ciphertext.buffer, ciphertext.byteOffset, ciphertext.length),
```
mpetrunic, line 81: buffer is subclass of Uint8Array so you shouldn't be required to create new one

src/noise.ts
```diff
@@ -49,10 +49,18 @@ export class Noise implements INoiseConnection {
     this.useNoisePipes = false
 
     if (staticNoiseKey) {
-      const publicKey = x25519.publicKeyCreate(staticNoiseKey)
+      const keyPair = x25519.generateKeyPairFromSeed(staticNoiseKey)
```
mpetrunic, line 52: this seems wrong, given key is privateKey not seed imo

mpetrun5: Its called generateKeyPairFromSeed but it actually accepts SECRET_KEY_LENGTH of 32 which is not seed but actually the 'staticNoiseKey'. 
Maybe I misread something?

mpetrunic: you are right, weird naming. Just add code comment as to not confuse other people

mpetrunic (approved):

vasco-santos:
🎉 

Do you have any metrics on the bundle size difference?

vasco-santos:
~27% decrease on its bundle size, from 191.5kb to 141.2kb (minified+gzipped)

🚀 ❤️

mpetrunic:
> ~27% decrease on its bundle size, from 191.5kb to 141.2kb (minified+gzipped)
> 
> 🚀 ❤️

from our benchmark there is also speed improvment:
```
NEW:

Initializing handshake benchmark
Init complete, running benchmark
handshake x 39.97 ops/sec ±4.98% (67 runs sampled)
Done in 6.04s.


OLD:

Initializing handshake benchmark
Init complete, running benchmark
handshake x 27.23 ops/sec ±4.57% (67 runs sampled)
Done in 6.10s.
```

PR #103, opened 2021-07-07
chore: update peer-id dep

achingbrain:
There's a new release of the `peer-id` module that uses the `CID` class from the `multiformats` module instead of from `cids`.

This PR upgrades to the new version (also upgrades mocha and sinon too).

BREAKING CHANGE: the new `peer-id` module uses the new `CID` class and not the old one

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,8 +42,8 @@
     "chai": "^4.2.0",
     "events": "^3.2.0",
     "microtime": "^3.0.0",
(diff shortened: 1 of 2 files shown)


mpetrunic (approved):
This affects noise api, so I guess I should release major version?

achingbrain:
That would be safest, yes please.

PR #113, opened 2021-11-26
chore: update libp2p-crypto

achingbrain:
libp2p-crypto@0.20.0 uses noble-secp256k1 and noble-ed25519 which
are smaller and faster than the version in https://www.npmjs.com/package/secp256k1

This PR updates the dep so we don't pull in 0.19.x and 0.20.x in
consuming bundles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,8 +56,8 @@
     "it-pair": "^1.0.0",
     "it-pb-rpc": "^0.1.11",
     "it-pipe": "^1.1.0",
-    "libp2p-crypto": "^0.19.7",
-    "peer-id": "^0.15.3",
+    "libp2p-crypto": "^0.20.0",
+    "peer-id": "^0.15.4",
     "protobufjs": "^6.11.2",
     "uint8arrays": "^3.0.0"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -4,7 +4,7 @@
 
 "@achingbrain/dependency-check@^4.1.0":
   version "4.1.0"
-  resolved "https://registry.yarnpkg.com/@achingbrain/dependency-check/-/dependency-check-4.1.0.tgz#331ce957118d4ea9a9789f5b67b971d4960254b4"
+  resolved "https://registry.npmjs.org/@achingbrain/dependency-check/-/dependency-check-4.1.0.tgz"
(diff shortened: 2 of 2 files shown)


mpetrunic (approved):

dapplion (approved):
Would be nice to benchmark and consider using native bindings for Lodestar at least

achingbrain:
@dapplion native bindings are much, much faster 😁 , though require node 15+ - https://github.com/libp2p/js-libp2p-crypto/pull/211

This is ok, I think, as officially libp2p supports LTS+Current which as of a couple of weeks ago is 16 and 17.

dapplion:
> @dapplion native bindings are much, much faster grin , though require node 15+ - [libp2p/js-libp2p-crypto#211](https://github.com/libp2p/js-libp2p-crypto/pull/211)
> 
> ~This is ok, I think, as officially [libp2p supports LTS+Current](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#supported-versions) which as of a couple of weeks ago is 16 and 17.~
> 
> Actually, scratch that, I updated the libp2p-crypto PR to fall back to the pure-js implementation if Ed25519 isn't supported.

Oh! Very interested! gossip encryption shows up in our Lodestar benchmarks. Do you have benchmarks comparing native bindings, current lib and noble's? We just bumped Lodestar to use NodeJS v16 by default so it shouldn't be an issue :+1:

dapplion:
> @dapplion native bindings are much, much faster grin , though require node 15+ - [libp2p/js-libp2p-crypto#211](https://github.com/libp2p/js-libp2p-crypto/pull/211)
> 
> ~This is ok, I think, as officially [libp2p supports LTS+Current](https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#supported-versions) which as of a couple of weeks ago is 16 and 17.~
> 
> Actually, scratch that, I updated the libp2p-crypto PR to fall back to the pure-js implementation if Ed25519 isn't supported.

Oh! Very interested! We just bumped Lodestar to use NodeJS v16 by default so it shouldn't be an issue :+1:

Damn with those benchmarks, crazy differences

achingbrain:
Superseded by #115

PR #115, opened 2021-11-30
chore: make libp2p-crypto a dev dep

achingbrain:
It's only used to verify incoming message payloads but the `pubKey` property on `PeerId` can do that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "buffer": "^6.0.3",
     "chai": "^4.3.4",
     "events": "^3.3.0",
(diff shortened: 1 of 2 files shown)


dapplion:
Makes sense! Agree with the change. However should we keep both the package-lock.json and yarn.lock?

achingbrain:
> should we keep both the package-lock.json and yarn.lock?

Ha! Oops.  I've removed the newly added `package-lock.json` file.

dapplion (approved):

PR #119, opened 2021-12-02
chore: bump version

wemeetagain:


vasco-santos (approved):

achingbrain:
This should be a major, the `peer-id` update breaks compatibility with node 14 and below.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@chainsafe/libp2p-noise",
-  "version": "4.1.1",
+  "version": "4.1.2",
```
achingbrain, line 3:
```suggestion
  "version": "5.0.0",
```

This should be a major, the `libp2p-crypto` update pulled in by the `peer-id` update breaks compatibility with node 14 and below.

npm will happily install it, but the `engines` field in `libp2p-crypto` was changed to `>=15.0.0` which will make yarn sad if not running the correct node version.

achingbrain (changes requested):

package.json
wemeetagain: Thanks for the review!
Good catch

achingbrain (approved):

tuyennhv (approved):

PR #120, opened 2021-12-06
chore: update changelog

wemeetagain:
releasing v5.0.0 failed because the changelog was not updated.
https://github.com/ChainSafe/js-libp2p-noise/runs/4433002833?check_suite_focus=true

We should consider having auto-generated changelogs for future releases


vasco-santos (approved):
LGTM

Let's create an issue for having an automated changelog

wemeetagain:
https://github.com/ChainSafe/js-libp2p-noise/issues/121

PR #125, opened 2022-01-03
fix: remove node buffers

achingbrain:
Replaces node buffers with browser-friendly `Uint8Array`s.

Previously we didn't want to do this because there's no equivalent to node's `Buffer.allocUnsafe` which doesn't set all entries of allocated buffers to 0.  The thing is we only actually use that in one place so to address this I've detected the existence of of `Buffer.allocUnsafe` before using it, otherwise falling back to `new Uint8Array` which is what the node `Buffer` polyfill does anyway.

Running the benchmark suite in this module shows the performance is comparable to or even slightly better than master (I think due to not having to convert `Uint8Array`s to `Buffer`s any more though please don't quote me on that 😆 ):

Before:

```console
$ node benchmark.js
Initializing handshake benchmark
Init complete, running benchmark
handshake x 59.95 ops/sec ±11.20% (75 runs sampled)
handshake x 54.68 ops/sec ±10.81% (68 runs sampled)
handshake x 50.42 ops/sec ±11.55% (65 runs sampled)
handshake x 53.41 ops/sec ±11.84% (68 runs sampled)
handshake x 50.25 ops/sec ±11.80% (66 runs sampled)
```

After:

```console
$ node ./benchmark.js
Initializing handshake benchmark
Init complete, running benchmark
handshake x 61.48 ops/sec ±11.71% (76 runs sampled)
handshake x 59.43 ops/sec ±11.13% (73 runs sampled)
handshake x 56.09 ops/sec ±12.02% (71 runs sampled)
handshake x 60.05 ops/sec ±11.69% (74 runs sampled)
handshake x 59.66 ops/sec ±10.59% (74 runs sampled)
```

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,27 +1,13 @@
 const path = require('path')
 
-/** @type {import('aegir').Options["build"]["config"]} */
-const esbuild = {
-  inject: [path.join(__dirname, 'test/fixtures/node-globals.js')]
-}
-
-
 /** @type {import('aegir').PartialOptions} */
 const config = {
     tsRepo: true,
     docs: {
       entryPoint: "src/index.ts"
     },
-    test: {
-      browser :{
-        config: {
-          buildConfig: esbuild
-        }
-      }
-    },
     build: {
-      bundlesizeMax: '231KB',
-      config: esbuild
+      bundlesizeMax: '231KB'
     }
 }
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,3 +172,5 @@ jobs:
             ${{ runner.os }}-yarn-
       - run: yarn --prefer-offline --frozen-lockfile
       - run: npx xvfb-maybe aegir test -t electron-renderer --bail
+        env:
+          ESBUILD_WORKER_THREADS: 0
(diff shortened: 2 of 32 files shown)


dapplion:
@achingbrain I've been benchmarking the memory usage of Uint8Array vs Buffer and Uint8Arrays take double memory to represent the same data. See benchmark results here https://github.com/ChainSafe/ssz/pull/219/files#diff-afd2490ed095916808da2e4d8c22e4c7ebc2062c020c722574214fdb908ef640

noise shouldn't persist too much data at once but it's an important consideration to be aware of

achingbrain:
That's interesting.  The percentage memory difference between `new Uint8Array(...)` and `Buffer.from(....)` seems to decrease as the size of the bytes increases so it may be some sort of fixed overhead, which is all weird as `Buffer` extends `Uint8Array`.  Maybe worth opening an issue against node itself to get some clarification?

As long as these things are getting garbage collected it shouldn't be a showstopper though?

wemeetagain (approved):
LGTM

dapplion:
@achingbrain Sorry forgot to go back to this

> As long as these things are getting garbage collected it shouldn't be a showstopper though?

I did a few heap snapshots and I can't see noise retaining any substantial amount of binary data. So good on this front :+1:

PR #129, opened 2022-01-14
fix: write handshake nonce correctly

achingbrain:
The `value` and `offset` args to `Buffer.writeUInt32LE` vs `DataView.setUint32` are the other way round.

Fixes a bug introduced in #125 where we were using the nonce value as an offset and writing `4` rather than writing the nonce value at offset `4`.

### DIFF
--- a/src/handshakes/abstract-handshake.ts
+++ b/src/handshakes/abstract-handshake.ts
@@ -60,7 +60,7 @@ export abstract class AbstractHandshake {
   protected nonceToBytes (n: uint64): bytes {
     // Even though we're treating the nonce as 8 bytes, RFC7539 specifies 12 bytes for a nonce.
     const nonce = new Uint8Array(12)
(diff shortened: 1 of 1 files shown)


mpetrunic:
I have no idea how you manage to pinpoint this but good job 👏🏻 👏🏻 👏🏻 👏🏻

mpetrunic (approved):

wemeetagain (approved):
Thanks ♥️

mpetrunic:
> @wemeetagain mind shipping a patch release with this? It seems we need it to fix CI: https://app.circleci.com/pipelines/github/ipfs/go-ipfs/6010/workflows/3f9c17d0-e388-43f8-8dee-0a7789c5928f/jobs/65407
> 
> ```
> failed to negotiate security protocol: error reading handshake message: chacha20poly1305: message authentication failed
> ```

Released [5.0.2](https://github.com/ChainSafe/js-libp2p-noise/releases/tag/v5.0.2)

PR #136, opened 2022-03-21
fix: access node process object via globalThis

achingbrain:
To allow use of noise in the browser with minimal bundling, access `process.env.FOO` via `globalThis` and with optional chaining.

Removes the need to use babel etc to transpile the code first.


mpetrunic (approved):

achingbrain:
@mpetrunic any chance of a release with this in it please?

dapplion:
@achingbrain Would you mind porting your auto-release infra to this library? Not time efficient to ask for releases on each timely fix

PR #140, opened 2022-04-11
fix: use protons and update project config

achingbrain:
- Swaps protobufjs for protons so we can generate `.ts` from `.proto` that is browser friendly (e.g. uses bigints instead of long.js, etc) and isn't broken by default
- Updates project config to add autorelease in the [unified CI model](https://github.com/protocol/.github) as requested in this comment - https://github.com/ChainSafe/js-libp2p-noise/pull/136#issuecomment-1075880511

Before merging this:

1. Check npm config of https://www.npmjs.com/package/@chainsafe/libp2p-noise - "Publishing access" should be set to "Require two-factor authentication or automation tokens"
2. Create a repository secret here called `NPM_TOKEN` with a npm automation token that can publish to `@chainsafe/libp2p-noise`.
3. Add https://github.com/web3-bot as a collaborator to this repo to receive automated CI config updates


achingbrain:
Yarn.

So, for some reason, `yarn@1.22.18` consistently symlinks the wrong `tsc` binary in `node_modules/.bin`.

The dep tree is:

```
@chainsafe/libp2p-noise@6.0.1 /Users/alex/Documents/Workspaces/ChainSafe/js-libp2p-noise
└─┬ aegir@37.0.12
.. other deps
  ├─┬ dependency-check@5.0.0-4
  │ └─┬ precinct@8.3.1
  │   └─┬ detective-typescript@7.0.2
  │     └── typescript@3.9.10
.. other deps
 └── typescript@4.6.3
```

It finds `typescript@3.9.10` as a dep of `detective-typescript` instead of `typescript@4.6.3` as a dep of `aegir`.

This is a problem because `typescript@3.9.10` doesn't support the right values for the [lib](https://www.typescriptlang.org/tsconfig#lib) option.

One solution might be to add a redundant `typescript` dep to this module and hope that the dependency resolution algorithm would choose that one to symlink, though it's a bad solution because it's not used directly here, plus you'd then need to keep it in sync with the `aegir` version which is extra work no-one needs.

[Unified CI](https://github.com/protocol/.github) uses npm and not yarn, so I've removed `yarn.lock` since it's redundant from a CI point of view but there's nothing to stop people using yarn if they want to, though at the moment it seems to have a bug in it.

achingbrain:
I have split this into two PRs - #141 and #142

PR #141, opened 2022-04-11
fix: use protons instead of protobuf for native .proto to .ts

achingbrain:
Swaps protobufjs for protons so we can generate `.ts` from `.proto` that is browser friendly (e.g. uses bigints instead of long.js, etc) and isn't broken by default.


achingbrain:
Interestingly this has the same problem with yarn as #142:

```console
% ls -l ./node_modules/.bin/tsc                
lrwxr-xr-x  1 alex  staff  55 11 Apr 12:00 ./node_modules/.bin/tsc -> ../detective-typescript/node_modules/typescript/bin/tsc
```

dapplion (approved):
LGTM!

mpetrunic (approved):

PR #153, opened 2022-05-23
feat: adjust to comply with conn encryptor spec

mpetrunic:
Errors now have additional `code` properties to match spec


mpetrunic:
cc @achingbrain

achingbrain (approved):

PR #159, opened 2022-06-17
chore!: update deps

achingbrain:
Updates all libp2p interfaces and test suites

BREAKING CHANGE: updates to the single-issue libp2p interface definitions


mpetrunic (approved):
was just about to do that locally, luckily got notification about PR 😄

PR #162, opened 2022-06-21
fix: improve nonce performance

tuyennhv:
**Motivation**
- There's a performance issue of `nonceToBytes()` as shown in https://github.com/ChainSafe/lodestar/issues/4140#issuecomment-1161174710
- Maintaining both Uint8Array and DataView and uint64 help improving nonce significantly

**Description**
- New `Nonce` class containing above data with `getUint64()`, `getBytes()` and `increase()` method
- New benchmark shows it's 150x faster

### DIFF
--- a/benchmarks/nonce.js
+++ b/benchmarks/nonce.js
@@ -1,0 +1,45 @@
+/* eslint-disable */
+import benchmark from 'benchmark'
+import { Nonce } from '../dist/src/@types/nonce.js'
+
+/**
+ * Using Nonce class is 150x faster than nonceToBytes
+ * nonceToBytes x 2.25 ops/sec ±1.41% (10 runs sampled)
+ * Nonce class x 341 ops/sec ±0.71% (87 runs sampled)
+ */
+function nonceToBytes (n) {
+  // Even though we're treating the nonce as 8 bytes, RFC7539 specifies 12 bytes for a nonce.
+  const nonce = new Uint8Array(12)
+  new DataView(nonce.buffer, nonce.byteOffset, nonce.byteLength).setUint32(4, n, true)
+  return nonce
+}
+const main = function () {
+  const bench1 = new benchmark('nonceToBytes', {
+    fn: function () {
+      for (let i = 1e6; i < 2 * 1e6; i++) {
+        nonceToBytes(i)
+      }
+    }
+  })
+    .on('complete', function (stats) {
+      console.log(String(stats.currentTarget))
+    })
+
+  bench1.run()
+
+  const bench2 = new benchmark('Nonce class', {
+    fn: function () {
+      const nonce = new Nonce(1e6)
+      for (let i = 1e6; i < 2 * 1e6; i++) {
+        nonce.increase()
+      }
+    }
+  })
+    .on('complete', function (stats) {
+      console.log(String(stats.currentTarget))
+    })
+
+  bench2.run()
+}
+
+main()
--- a/benchmarks/nonce.js
+++ b/benchmarks/nonce.js
@@ -1,6 +1,6 @@
 /* eslint-disable */
 import benchmark from 'benchmark'
-import { Nonce } from '../dist/src/@types/nonce.js'
+import { Nonce } from '../dist/src/nonce.js'
 
 /**
  * Using Nonce class is 150x faster than nonceToBytes
--- a/benchmarks/nonce.js
+++ b/benchmarks/nonce.js
@@ -31,7 +31,7 @@ const main = function () {
     fn: function () {
       const nonce = new Nonce(1e6)
       for (let i = 1e6; i < 2 * 1e6; i++) {
-        nonce.increase()
+        nonce.increment()
       }
     }
   })
--- a/src/@types/nonce.ts
+++ b/src/@types/nonce.ts
@@ -1,0 +1,32 @@
+import type { bytes, uint64 } from './basic'
+
+/**
+ * The nonce is an uint that's increased over time.
+ * Maintaining different representations help improve performance.
+ */
+export class Nonce {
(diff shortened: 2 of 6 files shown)


src/@types/nonce.ts
```diff
@@ -0,0 +1,32 @@
+import type { bytes, uint64 } from './basic'
+
+/**
+ * The nonce is an uint that's increased over time.
+ * Maintaining different representations help improve performance.
+ */
+export class Nonce {
```
mpetrunic, line 7: Don't think nonce impl should go in types dir

src/handshakes/abstract-handshake.ts
```diff
@@ -55,26 +52,13 @@ export abstract class AbstractHandshake {
...
-    // Even though we're treating the nonce as 8 bytes, RFC7539 specifies 12 bytes for a nonce.
-    const nonce = new Uint8Array(12)
-    new DataView(nonce.buffer, nonce.byteOffset, nonce.byteLength).setUint32(4, n, true)
-
-    return nonce
-  }
-
-  protected encrypt (k: bytes32, n: uint64, ad: Uint8Array, plaintext: Uint8Array): bytes {
-    if (n > MAX_NONCE) {
+  protected encrypt (cs: CipherState, ad: Uint8Array, plaintext: Uint8Array): bytes {
```
mpetrunic, line 55: why the interface change?

tuyennhv: will revert it back to pass in `k` and `n` 👍

mpetrunic:
Great to hear we will optimize noise 👍🏻

src/nonce.ts
```diff
@@ -0,0 +1,32 @@
...
+  private readonly view: DataView
+
+  constructor (n: uint64) {
+    this.n = n
+    this.bytes = new Uint8Array(12)
+    this.view = new DataView(this.bytes.buffer, this.bytes.byteOffset, this.bytes.byteLength)
+    this.view.setUint32(4, n, true)
+  }
+
+  increase (): void {
```
wemeetagain, line 19: Can you call this `increment`?

src/nonce.ts
```diff
@@ -0,0 +1,32 @@
+import type { bytes, uint64 } from './@types/basic'
+
```
wemeetagain, line 2: Can you move MIN_NONCE, MAX_NONCE and ERR_MAX_NONCE here?

src/handshakes/abstract-handshake.ts
```diff
@@ -89,13 +72,12 @@ export abstract class AbstractHandshake {
     return ciphertext
   }
 
-  protected decrypt (k: bytes32, n: uint64, ad: bytes, ciphertext: bytes): {plaintext: bytes, valid: boolean} {
-    if (n > MAX_NONCE) {
+  protected decrypt (k: bytes32, n: Nonce, ad: bytes, ciphertext: bytes): {plaintext: bytes, valid: boolean} {
+    if (n.getUint64() > MAX_NONCE) {
       throw new Error(ERR_MAX_NONCE)
     }
```
wemeetagain, line 78: This can be moved into a method on `Nonce`.
```ts
assertNonceInBounds(): void {
  if (this.n > MAX_NONCE) {
    throw new Error(ERR_MAX_NONCE)
  }
}
```

src/handshakes/abstract-handshake.ts
```diff
@@ -1,10 +1,11 @@
...
 import { fromString as uint8ArrayFromString } from 'uint8arrays'
-import type { bytes, bytes32, uint64 } from '../@types/basic.js'
+import type { bytes, bytes32 } from '../@types/basic.js'
 import type { CipherState, MessageBuffer, SymmetricState } from '../@types/handshake.js'
 import type { ICryptoInterface } from '../crypto.js'
 import { logger } from '../logger.js'
+import { Nonce } from '../nonce.js'
 
 export const MIN_NONCE = 0
 // For performance reasons, the nonce is represented as a JS `number`
```
wemeetagain, line 11: It may be worth revisiting this comment now that we have a more performant Nonce implementation.

Can you try changing `Nonce.n` to a `bigint`, changing MAX_NONCE to `2 ** 64 - 2`, and the dataview set to `setBigUint64`?
If it's not extremely expensive we should do it, since its part of the noise spec.

src/nonce.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * The nonce is an uint that's increased over time.
+ * Maintaining different representations help improve performance.
+ */
+export class Nonce {
+  private n: uint64
+  private readonly bytes: bytes
+  private readonly view: DataView
+
+  constructor (n: uint64) {
```
wemeetagain, line 12: maybe can be changed to
```ts
constructor (n = MIN_NONCE) {
  ...
```

wemeetagain (changes requested):
General structure looks good, good performance boost

benchmarks/nonce.js
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable */
+import benchmark from 'benchmark'
```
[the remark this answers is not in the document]
tuyennhv: 👍 I'll do it in another PR, there's another benchmark to migrate too

src/handshakes/abstract-handshake.ts
wemeetagain: ouch, yea lets just stick with `number`

I think there's a lingering bug, where the MAX_NONCE doesn't match the overflow from `dataview.setUint32`.
We might either change MAX_NONCE or do setUint32 twice for the high and low bytes.

mpetrunic (approved):

PR #165, opened 2022-07-13
chore: clean up logic

wemeetagain:
- decodePayload is synchronous
- remove isValidPeerId - only decode peer id once
- don't use the return value of verifySignedPayload anywhere
- remove duplicate length check, use isValidPublicKey


mpetrunic (approved):
We should probably add misuse promises rule to ipfs eslint config

PR #172, opened 2022-07-29
chore: update protons and peer collections

mpetrunic:
cc @achingbrain


wemeetagain (dismissed):

src/utils.ts
```diff
@@ -36,7 +36,7 @@ export function createHandshakePayload (
     identityKey: libp2pPublicKey,
     identitySig: signedPayload,
     data: earlyData ?? new Uint8Array(0)
-  })
+  }).slice()
```
achingbrain, line 39: `.slice()` is always a copy operation, `.subarray()` is a no-copy if it can be (e.g. the requested data is contained within one `Uint8Array` in the `Uint8ArrayList`).

`.sublist()` is better as it's always no-copy.  If you update `it-length-prefixed` and `it-pb-stream` to the latest versions in this PR, you should be able to pass `Uint8ArrayList`s all the way down without needing to convert to `Uint8Array`s.

```suggestion
  }).subarray()
```

achingbrain (changes requested):

src/utils.ts
mpetrunic: I think I had some type issues tring to update them, also do we plan on modifying encrypt interface to use uint8arraylist?

achingbrain: I just shipped the `it-pb-stream` update so that should fix any remaining problems.

> do we plan on modifying encrypt interface to use uint8arraylist?

I think we should, let's try to avoid copying memory wherever possible.  Can you open a PR to do that?

mpetrunic: yup

achingbrain (approved):

PR #177, opened 2022-08-02
chore!: remove noise pipes option

dapplion:
As discussed offline, not need to keep code that will probably be never used

BREAKING CHANGE: noise pipes option no longer supported


mpetrunic:
@dapplion lint errors^^

dapplion:
> @dapplion lint errors^^

So I had to disable prettier for this repo because it completely changed everything. But then there's this linter? So in Lodestar prettier is connected to the linter such that VSCode can understand what the linter wants and apply that automatically for you if you have  `{ "editor.formatOnSave": false }`. How do you handle that with aegir?

mpetrunic:
> > @dapplion lint errors^^
> 
> So I had to disable prettier for this repo because it completely changed everything. But then there's this linter? So in Lodestar prettier is connected to the linter such that VSCode can understand what the linter wants and apply that automatically for you if you have `{ "editor.formatOnSave": true }`. How do you handle that with aegir?

For me it detected. Do you have this checked: 
image

mpetrunic:
@dapplion since this is breaking change, should we wait with releasing until we land changed connection encrypted interface?

mpetrunic:
> @dapplion since this is breaking change, should we wait with releasing until we land changed connection encrypted interface?

@wemeetagain^^

wemeetagain:
yea lets wait, should happen soon

src/noise.ts
```diff
@@ -220,7 +145,7 @@ export class Noise implements INoiseConnection {
       await handshake.exchange()
       await handshake.finish()
 
-      if (this.useNoisePipes && handshake.remotePeer) {
+      if (handshake.remotePeer) {
```
mpetrunic, line 148: this condition and keycache store should be removed

mpetrunic (changes requested):

dapplion:
> > > @dapplion lint errors^^
> > 
> > 
> > So I had to disable prettier for this repo because it completely changed everything. But then there's this linter? So in Lodestar prettier is connected to the linter such that VSCode can understand what the linter wants and apply that automatically for you if you have `{ "editor.formatOnSave": true }`. How do you handle that with aegir?
> 
> For me it detected. Do you have this checked: image

Still not working, I can DM to debug

mpetrunic:
@achingbrain aegir dep-check is passing without warnings or errors for unused prod dependencies. Is that normal?

mpetrunic (approved):

Issue #181, opened 2022-08-05
Allow callers to optionally specify the prologue


John-LittleBearLabs:
While it's certainly nice to be allowed to let it default, it could also be useful to be _able_ to specify it.

This came up as I was working on an implementation for a new transport. I'd like to use this library, but the draft spec calls for a specific prologue. We could probably work around it downstream, but it seems cleaner and nicer to me to land a fix here that others can use.

mpetrunic:
resolved via https://github.com/ChainSafe/js-libp2p-noise/pull/182

mpetrunic:
:robot: Release is at https://github.com/ChainSafe/js-libp2p-noise/releases/tag/v8.0.0 :sunflower:

PR #182, opened 2022-08-05
fix: Allow the caller to set the prologue (#181)

John-LittleBearLabs:


src/noise.ts
```diff
@@ -32,7 +32,7 @@ export class Noise implements INoiseConnection {
   public protocol = '/noise'
   public crypto: ICryptoInterface
 
-  private readonly prologue = new Uint8Array(0)
+  private readonly prologue: Uint8Array;
```
achingbrain, line 35: Can you run the linter please, chances are it'll fail on the semicolon.

package.json
```diff
@@ -93,7 +93,8 @@
     "benchmark": "^2.1.4",
     "mkdirp": "^1.0.4",
     "protons": "^4.0.0",
-    "sinon": "^14.0.0"
+    "sinon": "^14.0.0",
+    "typescript": "^4.7.4"
```
achingbrain, line 97: `typescript` is not used by this repo directly, it's used by `aegir` which has it as a dependency so this shouldn't be necessary.

Are you using yarn?

achingbrain (changes requested):
LGTM, pending the comments.

Please can you add a test?

package.json
John-LittleBearLabs: I am using yarn. I was getting errors before this, and my hypothesis was that aegir was picking up my /usr/bin/tsc . 

I can remove this.

achingbrain: A dep that's a dep of aegir depends on an old ts version.  For some reason yarn chooses to symlink that one in `node_modules/.bin`, rather than the one aegir depends on, even though it's lower in the dep tree.

npm chooses the correct version to symlink 🤷

src/noise.ts
```diff
@@ -53,6 +53,11 @@ export class Noise implements INoiseConnection {
     } else {
       this.staticKeys = this.crypto.generateX25519KeyPair()
     }
+    if (prologueBytes) {
```
ckousik, line 56: nit: `this.prologue = prologueBytes ?? new Uint8Array(0);

mpetrunic:
@John-LittleBearLabs @ckousik Are you guys use noise pipes option (IK and XXFallback handshakes)?

src/noise.ts
John-LittleBearLabs: It did complain about semicolons, but not here.

````
/home/lbl/code/js-libp2p-noise/src/noise.ts:57:36: Extra semicolon. [Error/@typescript-eslint/semi]
/home/lbl/code/js-libp2p-noise/src/noise.ts:59:40: Extra semicolon. [Error/@typescript-eslint/semi]
````

In any case, I'm removing this semicolon since none of the other members next to it have semicolons. No reason for this one to be special (other than my editor formats-on-save).

ckousik:
> @John-LittleBearLabs @ckousik Are you guys use noise pipes option (IK and XXFallback handshakes)?

No, we are not using that option. We are using the XX handshake by default.

John-LittleBearLabs:
@achingbrain 

> Please can you add a test?

Added a test

wemeetagain:
@John-LittleBearLabs looks like linter is still complaining https://github.com/ChainSafe/js-libp2p-noise/runs/7732868911?check_suite_focus=true#step:5:11

John-LittleBearLabs:
@wemeetagain 

> @John-LittleBearLabs looks like linter is still complaining https://github.com/ChainSafe/js-libp2p-noise/runs/7732868911?check_suite_focus=true#step:5:11

Yeah sorry about that. Addressing it.

mpetrunic (approved):

Issue #180, opened 2022-08-04
How to avoid creating the default Noise instance?


alanshaw:
In Cloudflare workers you cannot call some crypto functions outside of a request:

```
Error: Some functionality, such as asynchronous I/O (fetch, Cache API, KV), timeouts (setTimeout, setInterval), and generating random values (crypto.getRandomValues, crypto.subtle.generateKey), can only be performed while handling a request.
```

This error occurs when importing the module because a default instance is created up front (the `NOISE` export).

I've considered a couple of solutions:

1. Change `const NOISE` to a `function NOISE()` that lazily creates (and caches) the default instance when called. This would be a BREAKING CHANGE for consumers (albeit an easy one to migrate to).
2. Add a package export for `"./noise": "./dist/src/noise.js"` so you can `import { Noise } from '@chainsafe/libp2p-noise/noise'` to avoid the creation of the default instance altogether. Not a breaking change but a bit clunky.

Let me know what you think or if you have any better solutions 🙂

alanshaw:
Any opinions?

mpetrunic:
I think we plan to release the breaking change very soon anyways and IMO we can remove that default instance.

mpetrunic:
:robot: Release is at https://github.com/ChainSafe/js-libp2p-noise/releases/tag/v8.0.0 :sunflower:

PR #188, opened 2022-08-09
feat!: remove default instance

alanshaw:
In Cloudflare workers you cannot call some crypto functions outside of a request:

```
Error: Some functionality, such as asynchronous I/O (fetch, Cache API, KV), timeouts (setTimeout, setInterval), and generating random values (crypto.getRandomValues, crypto.subtle.generateKey), can only be performed while handling a request.
```

This error occurs when importing the module because a default instance is created up front (the `NOISE` export).

This PR removes the default `NOISE` export to allow folks to create an instance when needed.

BREAKING CHANGE: The default instance `NOISE` is not longer created or exported. Import `Noise` and use `new Noise()` in place of `NOISE`.

resolves https://github.com/ChainSafe/js-libp2p-noise/issues/180

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,15 +25,14 @@ Install with `yarn add @chainsafe/libp2p-noise` or `npm i @chainsafe/libp2p-nois
 Example of using default noise configuration and passing it to the libp2p config:
 
 ```js
-import {NOISE, Noise} from "@chainsafe/libp2p-noise"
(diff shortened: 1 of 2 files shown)


wemeetagain (approved):
LGTM

mpetrunic (approved):

PR #202, opened 2022-09-01
feat: use allocUnsafe() where possible

tuyennhv:
**Motivation**
- allocUnsafe() is better in NodeJS environment
- there is an external memory spike issue in lodestar

**Description**
- Use `allocUnsafe` where possible

### DIFF
--- a/src/nonce.ts
+++ b/src/nonce.ts
@@ -1,5 +1,5 @@
 import type { bytes, uint64 } from './@types/basic'
-import { allocUnsafe } from './utils.js'
+import { allocUnsafe } from 'uint8arrays/alloc'
 
 export const MIN_NONCE = 0
 // For performance reasons, the nonce is represented as a JS `number`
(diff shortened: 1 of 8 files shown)


src/nonce.ts
```diff
@@ -22,7 +23,7 @@ export class Nonce {
 
   constructor (n = MIN_NONCE) {
     this.n = n
-    this.bytes = new Uint8Array(12)
+    this.bytes = allocUnsafe(12)
```
mpetrunic, line 26: this might be dangerous as it could already contain some data.

tuyennhv: we immediately set value below so I think it's safe for this place but you're right, will need to review all other places

tuyennhv: notice that in other dependencies of js-libp2p we do both `alloc()` and `allocUnsafe()` so we may want to do the same thing https://github.com/achingbrain/uint8arrays/blob/56329d16d6ca575c9638f3abb9601b8a034783b8/src/alloc.js#L8

wemeetagain:
We should use the `alloc` and `allocUnsafe` utils already included in `uint8arrays`

tuyennhv:
this was not proved to improve performance in lodestar, closing for now

PR #215, opened 2022-09-27
Add Noise Extensions and update deps

MarcoPolo:
* Updates some deps
* Adds the Noise Extensions registry

The Noise extensions is used by https://github.com/libp2p/js-libp2p-webtransport/


MarcoPolo:
cc @achingbrain since I'm not sure who to tag here for a review.

src/proto/payload.proto
```diff
@@ -1,8 +1,12 @@
...
-	bytes identity_sig = 2;
-	bytes data = 3;
+message NoiseExtensions {
+    repeated bytes webtransport_certhashes = 1;
 }
+
+message NoiseHandshakePayload {
+  optional bytes identity_key = 1;
+  optional bytes identity_sig = 2;
+  optional bytes data         = 3;
```
mpetrunic, line 10: According to https://github.com/libp2p/specs/pull/453/files this should also be removed

src/noise.ts
```diff
@@ -76,6 +77,8 @@ export class Noise implements INoiseConnection {
     return {
       conn,
       remoteEarlyData: handshake.remoteEarlyData,
```
mpetrunic, line 79: shouldn't this be also removed?

There also also other places where earlyData is referenced

src/proto/payload.proto
```diff
@@ -1,8 +1,12 @@
-syntax = "proto3";
-package pb;
+syntax = "proto2";
```
mpetrunic, line 1: please leave syntax to be proto3, protons library is not supporting proto2 officially and can sometimes generate weird results. In proto3 fields are optional by default so you can omit optional keywords.

mpetrunic (changes requested):
Thnx for the effort 👍🏻 
We are removing data/early data which is a breaking change so the title and description should be updated accordingly.

We should merge this after https://github.com/libp2p/js-libp2p-interfaces/pull/293 is released

src/proto/payload.proto
MarcoPolo: Yes, but I didn't want to make the breaking change here. This updates the api to make clients no longer able to send early data, but they can still receive early data.

src/proto/payload.proto
MarcoPolo: This is on purpose. The spec recently changed to go to proto2 since proto3 has some unwanted semantics: https://github.com/libp2p/specs/pull/456

src/noise.ts
MarcoPolo: Same comment as https://github.com/ChainSafe/js-libp2p-noise/pull/215#discussion_r982746339. Happy to do the breaking change here. Wdyt?

src/proto/payload.proto
mpetrunic: I understand that but it also says it's compatible with proto3. And I would rather not switch to proto2 because of and conclusion there: https://github.com/ipfs/protons/issues/34

src/noise.ts
MarcoPolo: I'll make this breaking change.

src/proto/payload.proto
achingbrain: > `optional NoiseExtensions extensions = 4;` is represented as a oneof in proto3

go-libp2p's older noise `.proto` file [with protobuf v3](https://github.com/libp2p/go-libp2p/blob/a82be49036096c93f44287780a8d32142b591976/p2p/security/noise/pb/payload.proto) doesn't have the `oneof` keyword.  I don't think it's implicit if not set so I'm not sure this is correct.  Am I missing something?

> but it's a bit of a hack.

Is this because the field identifier is non-sequential?  I think this is a feature of protocol buffers, any non-defined field should be skipped by the parser so this seems ok.

MarcoPolo: > > `optional NoiseExtensions extensions = 4;` is represented as a oneof in proto3
> 
> go-libp2p's older noise `.proto` file [with protobuf v3](https://github.com/libp2p/go-libp2p/blob/a82be49036096c93f44287780a8d32142b591976/p2p/security/noise/pb/payload.proto) doesn't have the `oneof` keyword. I don't think it's implicit if not set so I'm not sure this is correct. Am I missing something?

Sorry I didn't explain this fully. I'll explain more in a bit 


> > but it's a bit of a hack.
> 
> Is this because the field identifier is non-sequential? I think this is a feature of protocol buffers, any non-defined field should be skipped by the parser so this seems ok.

It's a hack because the real definition is proto2 but we're using a proto3 definition to workaround a protobuf-compiler bug.

achingbrain: > In proto3 fields are optional by default so you can omit optional keywords.

@mpetrunic They're `singluar` by default so we need the `optional` keyword - https://developers.google.com/protocol-buffers/docs/proto3#specifying_field_rules

MarcoPolo: These two protobufs generate the same exact code (which is probably a separate bug since a proto2 encoder _should_ write empty bytes if set. See https://github.com/libp2p/specs/issues/465 for a discussion around proto2 vs proto3):

```
syntax = "proto2";

message NoiseExtensions {
    repeated bytes webtransport_certhashes = 1;
}

message NoiseHandshakePayload {
  optional bytes identity_key = 1;
  optional bytes identity_sig = 2;
  optional NoiseExtensions extensions = 4;
}
```

```
syntax = "proto3";

message NoiseExtensions {
    repeated bytes webtransport_certhashes = 1;
}

message NoiseHandshakePayload {
  bytes identity_key = 1;
  bytes identity_sig = 2;
  NoiseExtensions extensions = 4;
}
```

---

I think it may make sense to use proto3 long term here, but we still need to discuss that. For the mean time, I can set this as a proto3 definition that I've manually validated is compatible with proto2 parsers: https://github.com/MarcoPolo/proto2and3-playground/blob/9eb3987c04b348d64d825104813ad27c4df691e7/src/main.rs#L14

MarcoPolo:
> @MarcoPolo interfaces are released so when you get a chance please update PR.

Rebased

MarcoPolo:
Let me fix the ts-expect-errors

MarcoPolo:
@mpetrunic Ready again :)

mpetrunic:
@MarcoPolo Had to mark extensions as optional in the proto file otherwise interop tests with the old proto file would fail with "Protocol error: value for required field "extensions" was not found in protobuf".

@achingbrain But according to the proto3 language guide this shouldn't happen for singular fields, right? Would that mean that protons is deserializing incorrectly?

MarcoPolo:
> @MarcoPolo Had to mark extensions as optional in the proto file otherwise interop tests with the old proto file would fail with "Protocol error: value for required field "extensions" was not found in protobuf".

The interop test is probably wrong because extensions was never required and proto3 doesn't have a notion of required fields. 


> @achingbrain But according to the proto3 language guide this shouldn't happen for singular fields, right? 

Correct. 

> Would that mean that protons is deserializing incorrectly?

I'm not sure. I'd have to look into it a bit closer. Could you link the failing test please?

MarcoPolo:
It should be safe to mark extensions as optional since either way explicit presence is tracked: https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#presence-in-proto3-apis.

mpetrunic:
> I'm not sure. I'd have to look into it a bit closer. Could you link the failing test please?

https://github.com/ChainSafe/js-libp2p-noise/blob/master/test/interop.ts

I'm fine with merging this, but it would be nice to get @achingbrain input on why protons error on missing singular field

PR #224, opened 2022-10-08
Add metrics

dapplion:
Minimal set of metrics to track errors and measure packet rate

- xxHandshakeSuccesses: Total count of xxHandshake successes
- xxHandshakeErrors: Total count of xxHandshake errors
- encryptedPackets: Total count of successfully encrypted packets
- decryptedPackets: Total count of successfully decrypted packets
- decryptErrors: Total count of errors decrypting packets


achingbrain (changes requested):
Can these metrics please get added to the [built in libp2p metrics object](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-metrics/src/index.ts#L192) instead of being their own thing?

dapplion:
`ComponentMetricsUpdate` is quite strange, how does this connects to Prometheus?

dapplion:
Updated PR to use a "metrics generator" so that consumer only needs to pass something that can create Gauge instances. This follows similar encapsulation principles as [built in libp2p metrics object](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-metrics/src/index.ts#L192) but with no performance penalty.

PR #229, opened 2022-10-12
fix!: remove @libp2p/components

achingbrain:
`@libp2p/components` is a choke-point for our dependency graph as it depends on every interface, meaning when one interface revs a major `@libp2p/components` major has to change too which means every module depending on it also needs a major.

Switch instead to constructor injection of simple objects that let modules declare their dependencies on interfaces directly instead of indirectly via `@libp2p/components`

Refs https://github.com/libp2p/js-libp2p-components/issues/6

BREAKING CHANGE: modules no longer implement `Initializable` instead switching to constructor injection


dapplion:
The changes of this PR do not remove `@libp2p/components` since is not used in noise, or am I missing something?

The changes of this PR appear to be:
- Bump protons, uint8arrays
- change Noise constructor to take an object (this appears to be a BREAKING CHANGE?)
- Rename `writer` variable to `w`
- Change `src/proto/payload.ts` logic

achingbrain:
It's part of a larger suite of changes to switch to constructor based injection for all libp2p components - see the linked issue in https://github.com/libp2p/js-libp2p-components/issues/6

* Bump protons, uint8arrays
  * yes all deps up to date
* change Noise constructor to take an object (this appears to be a BREAKING CHANGE?)
  * see above about constructor injection
* Rename writer variable to w
  * this is part of the protons update
* Change src/proto/payload.ts logic
  * this is part of the protons update to be more conformant with other protobuf implementations - see:
    * https://github.com/ipfs/protons/issues/43
    * https://github.com/ipfs/protons/pull/66

dapplion (approved):

achingbrain:
Could this be merged if everyone's happy?  `libp2p@0.40.x` is out and this is needed to roll it all up to `ipfs`.

dapplion (approved):
LGTM @wemeetagain @mpetrunic please review

wemeetagain (approved):

PR #236, opened 2022-10-14
chore: Add prefix in encryptStream

dapplion:
Javascript abstractions are not free. While using pipe here looks nice, it adds a non-neglible cost allocating Promises for each extra `for await ()` iteration.

- Similar rationale to https://github.com/libp2p/js-libp2p/issues/1420#issuecomment-1273272662

Move code to add prefix into encryptStream. I'll attempt the same for decryptStream but it's a bit more complex so prefer to split in two PRs

### DIFF
--- a/src/encoder.ts
+++ b/src/encoder.ts
@@ -2,7 +2,7 @@ import { concat as uint8ArrayConcat } from 'uint8arrays/concat'
 import type { Uint8ArrayList } from 'uint8arraylist'
 import type { bytes } from './@types/basic.js'
 import type { MessageBuffer } from './@types/handshake.js'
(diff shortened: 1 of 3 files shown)


src/encoder.ts
```diff
@@ -12,7 +12,7 @@ const allocUnsafe = (len: number): Uint8Array => {
   return new Uint8Array(len)
 }
 
-export const uint16BEEncode: LengthEncoderFunction = (value: number) => {
+export const uint16BEEncode = (value: number): Uint8Array => {
```
mpetrunic, line 15: why is the type changed here?

mpetrunic:
Also there are some lint issues^^

wemeetagain (approved):

PR #241, opened 2022-10-21
fix: only increment nonce on valid messages

wemeetagain:
See issue 2 here: https://github.com/flynn/noise/security/advisories/GHSA-g9mp-8g3h-3c5c


mpetrunic:
Already resolved via #200

wemeetagain:
lol i must have amnesia

mpetrunic:
I was laughing so hard when I have seen merge conflict between your new and old code 🤣

PR #242, opened 2022-10-25
feat: reuse encrypted data to avoid memory allocation

tuyennhv:
**Motivation**
As tested in lodestar, memory allocation is expensive for `decrypt()` flow

Screen Shot 2022-10-25 at 11 13 58

**Description**
- Stablelib already supports option `dst?: Uint8Array` parameter in its chacha20poly1305 implementation, the new `as-chacha20poly1305` should follow the same interface
- When decrypting, reuse the same memory as encrypted message is not used after that

### DIFF
--- a/src/crypto/streaming.ts
+++ b/src/crypto/streaming.ts
@@ -1,3 +1,4 @@
+import { TAG_LENGTH } from '@stablelib/chacha20poly1305'
 import type { Transform } from 'it-stream-types'
 import type { Uint8ArrayList } from 'uint8arraylist'
 import type { IHandshake } from '../@types/handshake-interface.js'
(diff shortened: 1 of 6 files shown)


src/crypto/streaming.ts
```diff
@@ -30,7 +31,13 @@ export function decryptStream (handshake: IHandshake): Transform<Uint8ArrayList,
...
         }
 
-        const { plaintext: decrypted, valid } = await handshake.decrypt(chunk.subarray(i, end), handshake.session)
+        if (end - TAG_LENGTH < i) {
+          throw new Error('Invalid chunk')
+        }
+        const encrypted = chunk.subarray(i, end)
+        // reuse the encrypted Uint8Array to avoid memory allocation
+        const dst = chunk.subarray(i, end - TAG_LENGTH)
+        const { plaintext: decrypted, valid } = await handshake.decrypt(encrypted, handshake.session, dst)
```
wemeetagain, line 40: I get that this is the main feature of this PR.
This seems strange though, basically using the same buffer as both the input and the output.
It seems that by writing where we're reading, the input could be modified before/during processing?

tuyennhv: @wemeetagain  yeah agree that this looks strange, but if this is proved to be correct, for the performance reason (as shared in the PR description) I think we can go with it.

> It seems that by writing where we're reading, the input could be modified before/during processing?

no, input is read byte by byte 1 time and immediately set to output, see https://github.com/StableLib/stablelib/blob/a89a438fcbf855de6b2e9faa2630f03c3f3b3a54/packages/chacha/chacha.ts#L174

in `assemblyscript` implementation, I have a lot of unit tests for it https://github.com/ChainSafe/as-chacha20poly1305/pull/1/files#diff-25252846b58979dcaf4e41d47b3eadd7e4f335e7fb98da6c049b1f9cd011f381R48

I will add more comments for the context

tuyennhv:
Testing this PR along with `as-chacha20poly1305` on some lodestar nodes to confirm the correctness of this work

tuyennhv:
> Testing this PR along with `as-chacha20poly1305` on some lodestar nodes to confirm the correctness of this work

@wemeetagain  I found no issue with this branch after testing on some lodestar nodes, see https://github.com/ChainSafe/lodestar/pull/4787#issue-1457218947

wemeetagain (approved):
LGTM. Thanks for digging into this one.

PR #253, opened 2022-11-29
fix: readme usage example

### LINKED ISSUE #252: Usage example doesn't work because `Noise` class isn't exported
https://github.com/ChainSafe/js-libp2p-noise#usage

```
import {Noise} from "@chainsafe/libp2p-noise"
```

but this class isn't imported by index.ts: https://github.com/ChainSafe/js-libp2p-noise/blob/309ec0fae94b52d0b6e92b37c63a2a969d488675/src/index.ts#L2

TimDaub:
- Fixes #252


mpetrunic (approved):
Nice catch, tnx!

PR #257, opened 2023-01-06
fix!: update multiformats and related dependencies

achingbrain:
`multiformats@11.x.x` shipped with a [breaking change](https://github.com/multiformats/js-multiformats/pull/230) so update all deps using multiformats to the latest version.

The code changed is not used by this module so there's not a lot of changes here, but this should still go out as a major because the interface versions have changed.


test/interop.ts
```diff
@@ -72,6 +72,7 @@ async function createJsPeer (options: SpawnOptions): Promise<Daemon> {
   }
 
   const opts: Libp2pOptions = {
+    // @ts-expect-error @libp2p/interface-peer-id types are not aligned
```
achingbrain, line 75: This can be removed once `libp2p@0.42.x` ships

test/index.spec.ts
```diff
@@ -43,7 +44,7 @@ describe('Index', () => {
     const wrappedInbound = pbStream(inbound.conn)
     const wrappedOutbound = pbStream(outbound.conn)
 
-    wrappedOutbound.writeLP(Buffer.from('test'))
+    wrappedOutbound.writeLP(uint8ArrayFromString('test'))
```
achingbrain, line 47: This was necessary for the browser tests to pass for me as `Buffer` is not in the browser global scope.

wemeetagain (approved):

PR #273, opened 2023-02-14
chore: Add peer id str to mismatch error

dapplion:
- Provide more context to debug mismatch instances like https://github.com/ChainSafe/lodestar/issues/4355

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -75,7 +75,7 @@ export async function verifySignedPayload (
   // Unmarshaling from PublicKey protobuf
   const payloadPeerId = await peerIdFromKeys(payload.identityKey)
   if (!payloadPeerId.equals(remotePeer)) {
-    throw new Error("Peer ID doesn't match libp2p public key.")
+    throw new Error(`Payload identity key ${payloadPeerId.toString()} does not match expected remote peer ${remotePeer.toString()}`)
   }
   const generatedPayload = getHandshakePayload(noiseStaticKey)
 
--- a/test/xx-handshake.spec.ts
+++ b/test/xx-handshake.spec.ts
@@ -88,7 +88,7 @@ describe('XX Handshake', () => {
       assert(false, 'Should throw exception')
     } catch (e) {
       const err = e as Error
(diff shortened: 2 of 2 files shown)


mpetrunic:
@dapplion Could you update test case expected error message^^

wemeetagain:
Really dislike tests like this, that check against specific string literals. Prone to breaking on updates and not much value add there.
Would be in favor of using `CodeError` and refactoring the test to check the err.code.

dapplion:
Cannot run tests locally, currently stuck building the library with

```
[21:35:54] tsc [started]
error TS2688: Cannot find type definition file for 'keyv'.
  The file is in the program because:
    Entry point for implicit type library 'keyv'


Found 1 error.

[21:35:55] tsc [failed]
[21:35:55] → Command failed with exit code 1: tsc
```

mpetrunic:
> Cannot run tests locally, currently stuck building the library with
> 
> ```
> [21:35:54] tsc [started]
> error TS2688: Cannot find type definition file for 'keyv'.
>   The file is in the program because:
>     Entry point for implicit type library 'keyv'
> 
> 
> Found 1 error.
> 
> [21:35:55] tsc [failed]
> [21:35:55] → Command failed with exit code 1: tsc
> ```

I cannot replicate that, have you tried deleting node_modules and package-lock.json?

dapplion:
removing package-lock.json did not fix the issue, but removing node_modules did. Why? Can we please commit package-lock.json to all of our repos? The DX is terrible

mpetrunic:
> removing package-lock.json did not fix the issue, but removing node_modules did. Why? Can we please commit package-lock.json to all of our repos? The DX is terrible. We never have this types of issues in the Lodestar monorepo

Not 100%, but I think because it prefers to use the local version instead of the latest one.
Solution should not be to remove node_modules but to bump the package in package.json to minimum acceptable version.

Putting lockfiles in the repo would indeed resolve this problem for us but that could still happen to users of js-libp2p-noise in case some other package they use depends on a specific version that we find acceptable even though it isn't.

mpetrunic (approved):

PR #316, opened 2023-04-19
fix!: update stream types

achingbrain:
libp2p streams are now explicit about the types of sync/sources they provide, showing that they are `AsyncGenerator`s and not just `AsyncIterable`s.

Refs: https://github.com/achingbrain/it-stream-types/pull/45

BREAKING CHANGE: the type of the source/sink properties have changed


wemeetagain:
@achingbrain it seems the interop test is failing

wemeetagain (dismissed):

mpetrunic:
@achingbrain Seems like go-libp2p in interop tests doesn't support mplex by default. I enabled yamux on js libp2p in this tests so it works but when debugging, go peer doesn't list mplex as supported protocol

mpetrunic (approved):

Issue #321, opened 2023-04-26
Security: Replace js crypto packages with noble


paulmillr:
Following dependencies:

https://github.com/ChainSafe/js-libp2p-noise/blob/de40989d0bd95db540680f0d30c8e2a10743a5ef/package.json#L79-L81

Can be replaced with **audited** https://github.com/paulmillr/noble-curves and https://github.com/paulmillr/noble-hashes which, I think, you're already using through `@libp2p/crypto`.

paulmillr:
Now also https://github.com/paulmillr/noble-ciphers, 1.6x faster than your current stablelib dep

PR #323, opened 2023-04-27
fix!: update hashing libs to @noble

mpetrunic:
related to #321 

BREAKING CHANGE: stablelib crypto backend renamed to pureJsCrypto


tuyennhv (approved):

dapplion:
@tuyennhv do we have a plan to keep using the fastest crypto backend available in Lodestar?

wemeetagain:
> we have a plan

There's no code change necessary in lodestar

PR #344, opened 2023-08-01
feat!: close streams gracefully

achingbrain:
- Updates all libp2p related deps
- Switches `it-pb-stream` for `it-length-prefixed-stream` and `it-byte-stream` since we only use the methods from those modules
- Stream close methods are now async

BREAKING CHANGE: stream close methods are now asyc, requires libp2p@0.46.x or later


achingbrain:
Blocked on https://github.com/ChainSafe/js-libp2p-yamux/pull/57

PR #385, opened 2023-11-10
fix: use native crypto if available

achingbrain:
Adds the lodestar [crypto
implementation](https://github.com/ChainSafe/lodestar/blob/6f27ac6e5f439577906cbfe5f9e01c59586c2af3/packages/beacon-node/src/network/libp2p/noise.ts#L3) to this module for use in node.

I'm trying to get js-libp2p back onto the [libp2p performance dashboard](https://observablehq.com/@libp2p-workspace/performance-dashboard) and in testing this change increases streaming throughput from around 60 MB/s to almost 300 MB/s.

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -1,0 +1,79 @@
+import crypto from 'node:crypto'
+import { newInstance, ChaCha20Poly1305 } from '@chainsafe/as-chacha20poly1305'
+import { digest } from '@chainsafe/as-sha256'
+import { pureJsCrypto } from './js.js'
+import type { ICryptoInterface } from '../crypto.js'
+
+const ctx = newInstance()
+const asImpl = new ChaCha20Poly1305(ctx)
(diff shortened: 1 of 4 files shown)


src/crypto/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    const final = cipher.final()
+    const tag = cipher.getAuthTag()
+
+    const encrypted = Buffer.concat([updated, tag, final])
+    return encrypted
+  },
+
+  chaCha20Poly1305Decrypt (ciphertext, nonce, ad, k, _dst) {
+    const authTag = ciphertext.slice(ciphertext.length - 16)
+    const text = ciphertext.slice(0, ciphertext.length - 16)
```
mpetrunic, line 31: can we use subarray instead?

achingbrain: Done!

mpetrunic (approved):

PR #388, opened 2023-11-22
fix: missing import extension

hackyguru:
This package is a dependency for https://github.com/waku-org/js-waku

But, due to the import on 'src/crypto/index.browser.ts' - `import { pureJsCrypto } from './.js'`, the CRA does not work. Fixed this by modifying the import

**Steps to reproduce**

1. Create a react app with `npx create-react-app <project-name>`
2. Install the waku sdk package from `npm install @waku/sdk`
3. Package installation fails and throws up the error because of the invalid import

This is a major issue because CRA will not along with package

### DIFF
--- a/src/crypto/index.browser.ts
+++ b/src/crypto/index.browser.ts
@@ -1,3 +1,3 @@
-import { pureJsCrypto } from './js'
+import { pureJsCrypto } from './js.js'
 
 export const defaultCrypto = pureJsCrypto


fryorcraken:
@hackyguru I think it would be good if you add the steps to reproduce the issue.

wemeetagain (approved):

achingbrain:
That's annoying. This module is imported during tests but because it's only used in browser testing it's run through esbuild first which I guess doesn't care about the missing extension?

mpetrunic:
> That's annoying. This module is imported during tests but because it's only used in browser testing it's run through esbuild first which I guess doesn't care about the missing extension?

Nope

PR #407, opened 2023-12-16
feat!: refactor codebase

wemeetagain:
Motivation:
- While debugging lodestar with js-libp2p v1.0 I was constantly visually tracing through this library.
- There were several organizational deficiencies that distracted from clearly understanding the code flow.
- This is likely because the usage of noise in libp2p has changed over time.
- This refactoring should result in better auditability of this library.

Review notes:
- Each commit _except the last one_ is small and digestible, can be reviewed individually
- Most changes come from the last commit
  - start with `src/protocol.ts` which contains the core noise protocol (agnostic to libp2p's usage)
  - then visit `src/performHandshake.ts` which uses the noise protocol according to the libp2p spec. The initiator and responder flows are broken into separate functions.
  - then everything else

Description
- Fixes to node crypto (in encryption and dh)
- Improved uint16 encode/decode perf
- Reduced allocations of known values (eg: the empty Uint8Array)
- Most types have been aligned with the noise spec where possible, better commented
- Replace code in `src/handshakes` with `src/protocol.ts`, this is noise protocol code, following the spec verbatim, agnostic to libp2p usage
- Replace code in `src/xx-handshake.ts with `src/performHandshake.ts` which performs a noise handshake, libp2p-style. This adds logging, reading/writing to a connection, and sending/receiving/verifying a libp2p noise payload.
- Utility functions surrounding the libp2p noise payload have been simplified 

BREAKING CHANGE:
The return type of `ICryptoInterface#chaCha20Poly1305Decrypt` has changed, now no longer returns `null` but rather throws on invalid decryption.


src/protocol.ts
```diff
@@ -0,0 +1,313 @@
...
+}
+
+export class CipherState implements ICipherState {
+  public k?: Uint8Array
+  public n: Nonce
+  private readonly crypto: ICrypto
+
+  constructor (crypto: ICrypto, k: Uint8Array | undefined = undefined, n = 0) {
+    this.crypto = crypto
+    this.k = k
```
maschad, line 24: Single letter variable names should at least have comments for clarity

src/crypto/index.ts
```diff
@@ -181,8 +185,7 @@ export const defaultCrypto: ICryptoInterface = {
         publicKey
       ], X25519_PREFIX.byteLength + publicKey.byteLength)
     } else {
-      publicKey.prepend(X25519_PREFIX)
-      publicKey = publicKey.subarray()
+      publicKey = new Uint8ArrayList(X25519_PREFIX, publicKey).subarray()
```
maschad, line 188: you should write a test to ensure that this state is consistent

wemeetagain: a regression test?
This fixes a bug where the input Uint8ArrayList was mutated. When that input Uint8ArrayList is part of the noise state, then that mutation breaks things.

src/protocol.ts
wemeetagain: If you use an IDE, all protocol-related types have comments.

src/crypto.ts
```diff
@@ -1,17 +1,27 @@
...
+  getHKDF(ck: Uint8Array, ikm: Uint8Array): [Uint8Array, Uint8Array, Uint8Array]
 
   generateX25519KeyPair(): KeyPair
   generateX25519KeyPairFromSeed(seed: Uint8Array): KeyPair
   generateX25519SharedKey(privateKey: Uint8Array | Uint8ArrayList, publicKey: Uint8Array | Uint8ArrayList): Uint8Array
 
-  chaCha20Poly1305Encrypt(plaintext: Uint8Array | Uint8ArrayList, nonce: Uint8Array, ad: Uint8Array, k: bytes32): Uint8ArrayList | Uint8Array
-  chaCha20Poly1305Decrypt(ciphertext: Uint8Array | Uint8ArrayList, nonce: Uint8Array, ad: Uint8Array, k: bytes32, dst?: Uint8Array): Uint8ArrayList | Uint8Array | null
+  chaCha20Poly1305Encrypt(plaintext: Uint8Array | Uint8ArrayList, nonce: Uint8Array, ad: Uint8Array, k: Uint8Array): Uint8ArrayList | Uint8Array
+  chaCha20Poly1305Decrypt(ciphertext: Uint8Array | Uint8ArrayList, nonce: Uint8Array, ad: Uint8Array, k: Uint8Array, dst?: Uint8Array): Uint8ArrayList | Uint8Array
```
mpetrunic, line 15: isn't this a breaking change? Should we release major then?

mpetrunic: Also should be added in BC section in PR description

wemeetagain: Ah right. will do

wemeetagain:
nevermind, it just appeared!

mpetrunic:
> nevermind, it just appeared!

Souch impatience and doubt in release-please 😋

PR #440, opened 2024-08-09
fix: make local peer id optional

achingbrain:
`libp2p@2.x.x` will remove the `localPeer` argument from the `secureInbound` and `secureOutbound` methods of the `ConnectionEncrypter` interface - see https://github.com/libp2p/js-libp2p/pull/2304.

Unfortunately the `js-libp2p` monorepo has a depdendency on this module so the change cannot be released until this module is compatible... with the unreleased change.

This PR tests the first argument to `secureInbound` and `secureOutbound` to ensure that it is actually a `PeerId`. If not it shuffles all the arguments along by one place.  This makes `@chainsafe/libp2p-noise` compatible with both `libp2p@1.x.x` and `libp2p@2.x.x`.

This PR can be reverted and the first argument removed once `libp2p@2.x.x` is released.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,11 +1,12 @@
 import { Noise } from './noise.js'
 import type { NoiseInit } from './noise.js'
 import type { NoiseExtensions } from './proto/payload.js'
-import type { ComponentLogger, ConnectionEncrypter, Metrics } from '@libp2p/interface'
+import type { ComponentLogger, ConnectionEncrypter, Metrics, PeerId } from '@libp2p/interface'
 export type { ICryptoInterface } from './crypto.js'
 export { pureJsCrypto } from './crypto/js.js'
 
 export interface NoiseComponents {
+  peerId: PeerId
   logger: ComponentLogger
   metrics?: Metrics
 }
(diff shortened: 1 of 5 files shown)


wemeetagain:
Won't there still be type errors since the interfaces don't match?

achingbrain:
I think there will be breakage after the `v2` release:

- `@chainsafe/libp2p-noise` currently resolves to the `@libp2p/interface` in the monorepo
- We have `"skipLibCheck": true` in the `js-libp2p` monorepo `tsconfig.json` so it doesn't check the types of `@chainsafe/libp2p-noise`
- When we release `v2`, `@chainsafe/libp2p-noise` will resolve to the older `@libp2p/interface` version
- At that point I think tsc will go 💥

achingbrain:
Maybe we'll need to do a major here shortly after the v2 release that updates the interface version, then a follow up that upgrades libp2p to the new noise major.

There will be a short window where things are broken but I'm not sure how else we can resolve the circular dependency, other than adding noise to the libp2p monorepo.

wemeetagain:
Can we cut a breaking release here that relies on a 2.0-compatible commit of libp2p/interface? then after 2.0 is released, rely on the official upstream again.

achingbrain:
Actually I wonder if we could use method overloading to be compatible with both versions of the interface?

wemeetagain:
> method overloading

worth a shot, should be able to be tested locally for build errors

wemeetagain (approved):
LGTM

PR #482, opened 2025-03-04
feat: select early muxer

achingbrain:
Implements early multiplexer negotiation to save multistream-select round trips when a connection is opened.

Spec: https://github.com/libp2p/specs/blob/6d38f88f7b2d16b0e4489298bcd0737a6d704f7e/connections/inlined-muxer-negotiation.md#multiplexer-negotiation-over-noise

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@noble/ciphers": "^1.1.3",
         "@noble/curves": "^1.1.0",
         "@noble/hashes": "^1.3.1",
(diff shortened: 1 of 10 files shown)


wemeetagain (approved):
Is this feature backwards compatible?

for example, if the client sends muxers, but the server doesn't recognize this feature yet, will the handshake succeed?

achingbrain:
Yes, if no muxer is agreed on using early data, the peers will fall back to using multistream-select to negotiate the muxer.

PR #608, opened 2025-08-06
fix: use scoped logger

achingbrain:
To better tie logging information to connections, use a scoped logger from the connection being secured to write log messages.


achingbrain:
Closing in favour of #618

PR #618, opened 2025-09-24
feat!: update to libp2p v3 api

achingbrain:
Refactors code to conform to the `libp2p@3.x.x` API.

- Encrypted streams are now EventTargets
- Consuming incoming data is now done synchronously for reduced latency

BREAKING CHANGE: Must be used with `libp2p@3.x.x`, it cannot be used with earlier versions


achingbrain:
Opened as a draft because https://github.com/ChainSafe/js-libp2p-yamux/pull/105 needs to ship first.

PR #665, opened 2026-04-04
feat: add post-quantum hybrid Noise handshake

paschal533:
## Post-quantum hybrid Noise handshake (XXhfs) for js-libp2p-noise

Implements `Noise_XXhfs_25519+MLKEM768_ChaChaPoly_SHA256`, the hybrid Noise handshake combining X25519 (classical) with ML-KEM-768 (post-quantum) in the KEM slot, as proposed in [libp2p/specs#723](https://github.com/libp2p/specs/issues/723). The wire format is specified in [libp2p/specs#727](https://github.com/libp2p/specs/pull/727), a Stage 1A Working Draft by @royzah. My own draft, [libp2p/specs#716](https://github.com/libp2p/specs/pull/716), was closed on 2026-09-18 in favour of #727 so that one document rather than two carries the proposal; its text stays readable at <https://github.com/paschal533/specs/tree/master/noise-pq>, and the material it covered (test vectors, the interoperability matrix, wire sizes and the downgrade analysis) was offered to #727.

**Status:** Research/WIP, not ready for merge, intended to demonstrate the integration point and validate cross-language interop.

**Breaking change (2026-09-17):** the suite was renamed from `Noise_XXhfs_25519+ML-KEM-768_ChaChaPoly_SHA256`, because Noise (revision 34, §8.2) allows only alphanumerics and `/` in algorithm names. The name is hashed into the handshake hash, so old-name and new-name peers cannot complete a handshake, and the protocol id moved from `/noise-mlkem768-hfs/0.1.0` to `/noise-mlkem768-hfs/0.2.0`. Message sizes are unchanged. `/noise-mlkem768-hfs/0.2.0` is the id this branch ships, not a spec-endorsed one: #727 writes `/noise-mlkem768-hfs/0.1.0` and lists the identifier string as the first of its open issues, so this will follow whatever #727 settles on. See the note in `NOISE_HFS_SPEC.md`.

### What this adds

- `src/crypto/pqc.ts`: ML-KEM-768 IKem backend via `@noble/post-quantum`
- `src/crypto/pqc.wasm.ts`: WASM backend slot (pure-JS stub until WASM is rebuilt for ML-KEM-768)
- `src/protocol-pqc.ts`: XXhfs handshake state machine wrapping the existing Noise XX machinery
- `src/noise-hfs.ts`: `NoiseHFS` ConnectionEncrypter, protocol ID `/noise-mlkem768-hfs/0.2.0`
- `test/pqc-kem.spec.ts`, `test/pqc-protocol.spec.ts`, `test/pqc-noise.spec.ts`, `test/pqc-vectors.spec.ts`: test suite, 99 tests, including replay of the five deterministic vectors in `test/fixtures/pqc-test-vectors.json`
- `scripts/node-listener.mjs` and `scripts/noise-hfs-dial.mjs`: TCP listener and dialer harnesses for cross-language interop testing; after the handshake each side sends one encrypted greeting

### Cross-language interop

On 2026-09-17 a neutral runner drove four implementations: this branch, [py-libp2p PR #1310](https://github.com/libp2p/py-libp2p/pull/1310) (Python), [nim-libp2p PR #2811](https://github.com/vacp2p/nim-libp2p/pull/2811) (Nim), and [rust-libp2p PR #6481](https://github.com/libp2p/rust-libp2p/pull/6481) by @royzah (Rust) with the harness from [royzah/rust-libp2p PR #1](https://github.com/royzah/rust-libp2p/pull/1). Every ordered listener/dialer pairing, each implementation against itself included, ran three times: **48 runs, 48 passed**. This branch was at `a183009`; the one later commit only corrects CHANGELOG.md.

| listener \ dialer | JS | Python | Nim | Rust |
|---|---|---|---|---|
| **JS** | 3/3 | 3/3 | 3/3 | 3/3 |
| **Python** | 3/3 | 3/3 | 3/3 | 3/3 |
| **Nim** | 3/3 | 3/3 | 3/3 | 3/3 |
| **Rust** | 3/3 | 3/3 | 3/3 | 3/3 |

A run passes only if both sides exit cleanly, each side reports the other's actual peer id, and one encrypted greeting goes each way, which exercises both `split()` cipher states. All runs were over loopback TCP on one Windows 11 machine. The harnesses start the handshake directly on TCP, so multistream-select negotiation of the protocol id is not covered. Results, `versions.txt` and all 96 logs: [pq-noise-artifacts, run 20260917T134954Z](https://github.com/paschal533/pq-noise-artifacts/tree/main/interop/results/20260917T134954Z). Negative controls (an old-name build that fails against every other implementation, and a fabricated identity that the cross-check catches): [interop/negative-controls](https://github.com/paschal533/pq-noise-artifacts/tree/main/interop/negative-controls).

**Earlier results, corrected.** An earlier version of this description showed a June 2026 table (Rust listener + Python dialer, Rust listener + JS dialer, Python listener + JS dialer, all PASS) on `/noise-mlkem768-hfs/0.1.0` with the hyphenated name. Those runs were handshake-only: the runner counted a pass when the dialer exited cleanly and printed a peer id, and no transport frames were exchanged. The Rust listener was ours, from royzah/rust-libp2p#1, not part of #6481, and its `Cargo.lock` pinned royzah's `snow` fork at a June commit that still used the hyphenated name. The matrix above replaces that table. Details: [artifacts README](https://github.com/paschal533/pq-noise-artifacts/blob/main/README.md).

### Key design decision

X-Wing (which bundles X25519 + ML-KEM-768 as a single KEM) was considered but rejected: the Noise XXhfs pattern already provides classical security through its own DH tokens (ee, es, se), making the X25519 inside X-Wing redundant. Raw ML-KEM-768 in the `ekem1` slot is the architecturally correct choice.

### ML-KEM-768 sizes

| Field | Size |
|-------|------|
| Public key (encapsulation key) | 1184 B |
| Secret key (decapsulation key) | 2400 B |
| Ciphertext | 1088 B |
| Shared secret | 32 B |
| Message A wire size | 1216 B (e=32 + e1=1184) |
| Encrypted ciphertext | 1104 B (1088 + 16 AEAD) |
| Message B wire size, empty payload | 1200 B |
| Message C wire size, empty payload | 64 B |


paschal533:
Quick update: I've put together a standalone example repo that shows the XXhfs handshake working end-to-end with real libp2p nodes. It's at https://github.com/paschal533/pqc-libp2p-example
 
The repo has two layers. The first is a pure-crypto walkthrough you can run with node scripts/handshake-walkthrough.js. It steps through the full XXhfs handshake in process with annotated output showing each message, the X-Wing encapsulate/decapsulate, and a KEM match confirmation. No libp2p involved, just the raw crypto.
 
The second is a full libp2p setup: a Node.js listener (npm run listener) and a browser client (npm run browser:dev) both negotiating `/noise-pq/1.0.0` over WebSockets. You paste the multiaddr into the browser UI, hit Connect, and it shows the remote peer ID, the protocol, and the handshake latency.

I'm also finalising a research paper covering the design rationale, security analysis, and benchmark numbers. I'll link it here once it's ready.

If anyone wants to try it out and has feedback, especially on the API surface or anything that looks off in the handshake walkthrough, I'd really appreciate it.

paschal533:
Ran the live cross-language interop test today, JS listener, Python dialer, real TCP connection.

**Setup:**
- Terminal 1: `cd js-libp2p-noise && node scripts/node-listener.mjs` (JS responder, port 8000)
- Terminal 2: `cd py-libp2p && python scripts/interop_dial.py` (Python initiator, libp2p/py-libp2p#1310)

**JS listener output:**

```
Listener peer ID: 12D3KooWHFFVYwAnVQHZTqZhcq1woX11xKsMnujNDPbDyh77kLBa
Protocol: /noise-pq/1.0.0

Listening on tcp://127.0.0.1:8000
Waiting for Python dialer...

Incoming TCP connection from 127.0.0.1:PORT
Starting NoiseHFS responder handshake...
Handshake complete! Remote peer: 12D3KooWRPBkhDbfQmRJjVpR7hkEAmu8FP6mEaTBW61BAryeNjAH
Sent: "hello from JS"
Received: "hello from Python"

✅ INTEROP SUCCESS: Both sides exchanged messages through NoiseHFS!
```

**Python dialer output:**

```
[interop_dial] TCP connection established
[interop_dial] Starting XXhfs handshake (Python = initiator)...
[interop_dial] Handshake complete! Remote peer: 12D3KooWHFFVYwAnVQHZTqZhcq1woX11xKsMnujNDPbDyh77kLBa
[interop_dial] Received from JS: "hello from JS"
[interop_dial] Sent to JS: "hello from Python"

INTEROP SUCCESS
Python <-> JavaScript NoiseHFS handshake complete.
Protocol: Noise_XXhfs_25519+XWing_ChaChaPoly_SHA256
```

Both runtimes independently derived the same session keys and successfully exchanged encrypted messages. The `scripts/node-listener.mjs` added in the latest commit is all you need on the JS side to reproduce this.

Node.js v22, Python 3.13, Windows 11.

paschal533:
## WASM backend results + Amdahl's Law finding

Following up on the performance numbers in the PR description. I built a Rust WASM module for the X-Wing KEM to see how much headroom there is on the KEM side, and the results turned up something interesting worth documenting here.

### What was built

A standalone Rust crate (`src-wasm/`) compiled with wasm-pack to a 58 KB `.wasm` binary. Stack: `ml-kem 0.3.0-rc.2` (RustCrypto FIPS 203), `x25519-dalek 2.0`, `sha3 0.10`, `wasm-bindgen 0.2`. Three exported functions: `xwing_keygen`, `xwing_encapsulate`, `xwing_decapsulate`. The secret key is stored as a 64-byte ML-KEM seed rather than the expanded 2400-byte decapsulation key, so the full keypair output is 1312 bytes instead of 3648.

### KEM micro-benchmark results (Node.js v22.17.1, x64)

| Operation | Pure-JS ms/op | WASM ms/op | Speedup |
|-----------|--------------|------------|---------|
| `generateKemKeyPair` | 3.42 | 1.40 | 2.4x |
| `encapsulate` | 8.32 | 2.19 | 3.8x |
| `decapsulate` | 7.33 | 3.10 | 2.4x |
| Full KEM round-trip | 21.43 | 6.72 | 3.2x |

The WASM numbers are stable across runs (less than 10% variance). The pure-JS numbers show more variance because of V8 JIT tier transitions on a busy dev machine.

### The surprising part: full handshake barely moves

| Protocol | ms/handshake |
|----------|-------------|
| Noise_XX classical | ~19 ms |
| Noise_XXhfs pure-JS KEM | ~93 ms |
| Noise_XXhfs WASM KEM | ~91 ms |

The WASM KEM is 3.2x faster but the full handshake only improves by about 2%. This is Amdahl's Law in action. The KEM is not the only thing happening in a handshake. A complete `XXhfs` exchange also does roughly:

- 8 SHA-256 invocations
- 4 HKDF derivations
- 4 ChaCha20-Poly1305 AEAD operations
- 2 X25519 DH computations
- 2 Ed25519 signature verifications
- 2 Protobuf encode/decode passes
- 6 async task boundaries (Promise.all, stream reads/writes)

All of those are still pure-JS on `pureJsCrypto` / `@noble/*`. The KEM accounts for maybe 25-30% of the total handshake time. Speeding up just that part can only recover a limited fraction of the overhead.

### What this means for the ~35-44 ms overhead

To actually get the full handshake close to classical, you need one of:

1. **Native Node.js ML-KEM support** (planned for v24+) - this would also help the SHA-256/ChaCha20 path since those are already native via `defaultCrypto`. Best case outcome.
2. **A full WASM crypto backend** covering all Noise primitives, not just the KEM. Significantly more work but achievable.
3. **WebCrypto ML-KEM** in browsers when it eventually lands.

For now the WASM module is a useful reference showing what the KEM-only ceiling looks like, and the 58 KB binary is small enough to ship without concern. But the big win on handshake latency is going to come from the Node.js native path, not from this.

The updated benchmark table and this analysis are now in the research write-up. Happy to add a note to `benchmarks/results.md` here too if that would be useful for reviewers.

paschal533:
Interop update, plus a performance result from another implementation that puts the numbers in this PR in context.

## Interop with the Nim implementation

I ran this branch against the Nim implementation of the same profile (vacp2p/nim-libp2p#2811) on 2026-09-05, in both roles:

- `scripts/noise-hfs-dial.mjs` dialling a Nim listener — handshake completed, peer id reported on both sides.
- `scripts/node-listener.mjs` listening, Nim dialling, with a message exchanged in each direction after the handshake.

That second one matters more than it looks. A completed handshake proves both sides agreed on the handshake hash and the KEM shared secret, but not that the two cipher states came out of `split()` with the same orientation — a swapped `cs1`/`cs2` still completes the handshake and only fails on the first data frame. Since `split()` assigns opposite states to initiator and responder, decrypting in each direction exercises a different key, so both are needed. Both worked.

With this and the June triangle test, all six pairwise combinations across the TypeScript, Python, Rust and Nim implementations have now completed a live handshake, none of them requiring a protocol change.

## One small fix pushed here

`scripts/node-listener.mjs` was checking the reply against the literal string `hello from Python`, so a perfectly good exchange with any other implementation was reported as an unexpected reply. It now matches the generic `hello from <impl>` form. That is the only change in this push.

## Performance context

I benchmarked the Nim implementation, which is the only one of the four whose KEM is compiled, assembly-optimised C (BoringSSL) rather than interpreted or WASM. Same machine, same protocol:

| Metric | This PR (JS) | Python | Nim |
|---|---:|---:|---:|
| Classical XX | 8.75 ms | 3.32 ms | 2.838 ms |
| XXhfs | 44.18 ms | 42.96 ms | 3.245 ms |
| Overhead | +5.0x | +12.9x | **+1.14x** |
| KEM share of handshake | ~48% | ~63% | ~9.9% |
| ML-KEM-768 round-trip | 21.43 ms | ~27 ms | **0.321 ms** |

The BoringSSL KEM round-trip is about 67x faster than `@noble/post-quantum` here and about 21x faster than the Rust WASM module I built earlier.

I think this supports the Amdahl argument in this PR rather than undercutting it. The WASM experiment showed that a 3.2x KEM speedup buys under 4% on the full handshake because the rest of the stack is still pure JS. The Nim result is the other end of that curve: once the KEM is genuinely fast, the KEM stops being the problem and the surrounding crypto is all that is left. For this implementation that points squarely at native Node.js ML-KEM in v24+ being the thing worth waiting for, and at a KEM-only optimisation not being worth much on its own.

Worth stating the caveat honestly: nim-libp2p optimises only the KEM — its X25519 and ChaCha20-Poly1305 come from BearSSL, SHA-256 from a pure-Nim library, Ed25519 from a pure-Nim reference port. So its 2.8 ms classical baseline is not a floor, and the 1.14x ratio is what that particular composition pays. Nobody has yet benchmarked this protocol with the whole Noise primitive set natively optimised.

paschal533:
I need to correct the performance numbers I posted earlier, and in doing so I found something in this PR that I think is worth fixing.

## The overhead figure was wrong, and it was my benchmark's fault

I reported +5.0x for the hybrid handshake. The real figure, measured properly, is **1.47x**. Two separate problems produced the difference.

**The benchmark was still measuring X-Wing.** `benchmarks/benchmark-pqc.js` kept hardcoded X-Wing wire constants after the June migration — `msgA: 32 + 1216`, `msgB: 32 + 1136` — reporting 2,352 B of overhead instead of 2,288 B, with labels to match. The handshake timings were always real ML-KEM-768 since they drive `noiseHFS()` directly, but the KEM microbenchmarks and the wire-size report were measuring a construction we no longer use. Fixed in 68189da; the wire report now derives 2,288 B, agreeing independently with the spec draft and the py-libp2p harness.

**`noise()` and `noiseHFS()` default to different crypto backends.** This is the bigger one.

```js
// noise.ts:51
const _crypto = crypto ?? defaultCrypto   // Node native + AssemblyScript WASM

// noise-hfs.ts:97
const _crypto = crypto ?? pureJsCrypto    // everything in JavaScript
```

So benchmarking one against the other measures the addition of a KEM *and* a wholesale switch of ChaCha20-Poly1305, SHA-256, X25519 and HKDF from native code to interpreted JavaScript — and attributes the sum to post-quantum cryptography.

I added `benchmarks/backend-isolation.mjs` to separate them. Medians of 30 iterations, four repetitions:

| Configuration | ms |
|---|---:|
| `Noise_XX` — native backend | 25.91 |
| `Noise_XX` — `pureJsCrypto` | 66.58 |
| `Noise_XXhfs` — `pureJsCrypto` + ML-KEM-768 | 79.07 |
| `Noise_XXhfs` — native backend + ML-KEM-768 | 38.12 |

The backend switch alone accounts for ~41 ms. The KEM alone costs ~12 ms — and ~12 ms on *either* backend, which is the sanity check that the decomposition is right, since `pqcKem` uses `@noble` regardless of which `ICryptoInterface` surrounds it.

Holding the backend constant: **1.47x, range 1.44–1.59 across four runs.** Not 4-5x.

## The suggestion

I don't think there's a reason for `noiseHFS()` to default to `pureJsCrypto`. `defaultCrypto` is browser-safe — `package.json`'s `browser` field swaps `crypto/index.js` for `crypto/index.browser.js`, which is exactly how classical `noise()` stays portable while using the faster path. As it stands the hybrid handshake pays roughly double on Node for nothing in return.

Changing the default looks like a one-line fix, but it's a behavioural change to this PR's public surface and I'd rather not make it unilaterally. Happy to push it if you agree it's right, or to leave it as an explicit opt-in if there's a reason for the current default I've missed.

## Also correcting myself

In my previous comment I said native Node.js ML-KEM in v24+ was "the thing worth waiting for". That was out of date when I wrote it — Node.js v24.7.0 shipped ML-KEM in WebCrypto in August 2025, along with ChaCha20-Poly1305 and SHA-3. `crypto.subtle` exposes ML-KEM-512/768/1024 plus `crypto.encapsulate()`/`decapsulate()`. So `src/crypto/pqc.node.ts` has something real to bind to now, and my figures above are still measured on v22.17.1, which means even the corrected 1.47x describes a runtime whose symmetric primitives are native but whose KEM is not.

## For context, the other implementations

Same protocol, measured the same week, each holding its own backend constant:

| | Python | JavaScript | Rust | Nim |
|---|---:|---:|---:|---:|
| KEM library | `kyber-py` | `@noble/post-quantum` | RustCrypto `ml-kem` | BoringSSL |
| Overhead | 10.5x | **1.47x** | 1.30x | 1.13x |

Three of the four sit between 1.1x and 1.5x. Only Python is an order of magnitude out, and only because `kyber-py` does the lattice arithmetic in pure Python.

One pattern I'll note because it seems worth naming: every measurement fault I found in this work pushed the same direction. The X-Wing leftovers, phase-separated sampling in three of the four harnesses, and this backend mismatch all made post-quantum cryptography look more expensive than it is. None made it look cheaper.

paschal533:
Pushed a breaking rename, a protocol id bump and a new cross-implementation interop run. Summary:

**Rename.** The suite is now `Noise_XXhfs_25519+MLKEM768_ChaChaPoly_SHA256`. Noise (revision 34, §8.2) allows only alphanumerics and `/` in an algorithm name, so `ML-KEM-768` was never valid. @royzah spotted it. It isn't cosmetic. The name is hashed into the handshake hash `h` (`InitializeSymmetric`, §5.2), which also seeds the chaining key, so old-name and new-name peers derive different keys. Message A goes in the clear and is byte-identical. Everything from the encrypted part of message B onwards changes. Message sizes don't change (1,216 / 1,200 / 64 B). `test/fixtures/pqc-test-vectors.json` is regenerated, and the 99 PQC tests pass. See the note in [`NOISE_HFS_SPEC.md`][js-spec].

**Protocol id.** `NoiseHFS` now uses `/noise-mlkem768-hfs/0.2.0`. Two builds that can't complete a handshake with each other shouldn't share an id: on 0.1.0 they'd negotiate and then fail decrypting message B, while with separate ids the mismatch shows up at negotiation instead. None of this is merged anywhere, so the bump costs nothing now. (The interop harnesses below start the handshake directly on TCP, so negotiation itself isn't exercised by them.)

**Interop, 48/48.** A neutral runner now drives all four implementations: this branch, libp2p/py-libp2p#1310, vacp2p/nim-libp2p#2811, and @royzah's libp2p/rust-libp2p#6481 with the harness from royzah/rust-libp2p#1. It runs every ordered listener/dialer pairing, each implementation against itself included, three times each, on 2026-09-17 with this branch at `a183009` (the one later commit only corrects CHANGELOG.md). A run passes only if both sides exit cleanly, each side reports the other's actual peer id, and one encrypted greeting goes each way. That last check exercises both `split()` cipher states, not just the handshake. `scripts/node-listener.mjs` and `scripts/noise-hfs-dial.mjs` are the harnesses it runs. Everything ran over loopback TCP on one Windows 11 machine. Results and all 96 logs: [run 20260917T134954Z][results].

**Negative controls.** This branch rebuilt with only the protocol name set back to the hyphenated form fails all six orderings against Python, Nim and Rust, with the dialer hitting an AEAD error on message B. Same-implementation pairs still pass ([control A][ctl-a], re-run against the revisions above). A dialer that prints a fake peer id completes the handshake and the greeting, and the identity cross-check catches it ([control B][ctl-b], from the earlier run `20260917T015709Z`).

**Correction.** The June 2026 Rust and Python pairings in the description, which my 5 September comment also leans on ("the June triangle test"), were handshake-only. The runner then counted a pass when the dialer exited cleanly and printed a peer id, and no transport frames were exchanged. The Rust side was our listener from royzah/rust-libp2p#1, not part of #6481, and its `Cargo.lock` pinned royzah's `snow` fork at a June commit that still used the hyphenated name. So those runs didn't test royzah's current code. The run above replaces them. Details: [artifacts README][readme].

**Benchmarks.** 2026-09-17 re-run on the renamed suite: [SUMMARY.md][bench]. The like-for-like overhead (native backend on both sides) is 1.56x. That's the median of the 5 pass-level medians, each the median of 30 per-iteration ratios, with a range of 1.51–1.60x. Absolute latencies from that session aren't comparable with earlier sessions. The file explains why. It lists earlier published figures for reference only and computes no ratio or delta against them.

[js-spec]: https://github.com/paschal533/js-libp2p-noise/blob/feat/pqc-xxhfs-noise/NOISE_HFS_SPEC.md
[results]: https://github.com/paschal533/pq-noise-artifacts/tree/main/interop/results/20260917T134954Z
[ctl-a]: https://github.com/paschal533/pq-noise-artifacts/blob/main/interop/negative-controls/20260917T134954Z/README.md
[ctl-b]: https://github.com/paschal533/pq-noise-artifacts/blob/main/interop/negative-controls/20260917/README.md
[readme]: https://github.com/paschal533/pq-noise-artifacts/blob/main/README.md
[bench]: https://github.com/paschal533/pq-noise-artifacts/blob/main/benchmarks/2026-09-17/SUMMARY.md

PR #666, opened 2026-04-16
deps: update protons to 9.x.x

achingbrain:
Updates to the latest protons and regenerates protobuf code.

This de-duplicates protons-runtime deps in browser bundles since the rest of the ecosystem has updated.


achingbrain:
Had to add a mocha dev dep because pnpm resolves dependencies differently to npm 😩

